PR #1564, opened 2014-12-29
[wrt][XWALK-2971] Add 1 new test case for Package Management

linshen:


linshen:
I check all WRT test cases, all test cases are already use BSD license, In addition,I would not use [master] prefix in the commit message again ,and I would use  test case instead of TCs, so I close this PR

ibelem:
@linshen  Please use the 2015 in BSD license.

ibelem:
LGTM

PR #1570, opened 2014-12-30
[wrt][XWALK-1499] Add 1 new test case

linshen:
Impacted TCs num(approved):New 1,Update 0,Delete 0
Test Result: pass 1, fail 0, block 0


linshen:
The XWALK-1499 is Won't Fix,So I close this PR

PR #1573, opened 2014-12-31
[wrt] hange shell scripts license to BSD

linshen:
These scripts were orignially created in 2013
checking by 'git log --follow'


linshen:
I check all WRT test cases, if changing author to authors does not make sense,then all test cases are already  use BSD license,see https://github.com/crosswalk-project/crosswalk-test-suite/pull/1564.so i close this PR

PR #1596, opened 2015-01-06
[usecase] Remove usecase for picture element

kangxu:


zqzhang:
Prefix bug id, https://crosswalk-project.org/jira/browse/XWALK-2999

kangxu:
@zqzhang  , I have updated it following your comments

PR #1598, opened 2015-01-06
[webapi] Update some case to the latest as W3C

chenxix:
Impacted tests(approved):new 0,update 10,delete 0
Unit test platform:[android]
Unit test result summary:pass 10,fail 0,block 0


zqzhang:
Why not submit to W3C firstly? Why these tests need `_requireManualCheck()`?

chenxix:
At the beginning when we get these cases from W3C,  _requireManualCheck() has been used already. And now canvas-tests.js script and these cases in W3C have been update, we just keep pace with them.

zqzhang:
Then this is to sync the test code to latest W3C repo, right? If yes, please update the commit message.

PR #1602, opened 2015-01-07
[wrt][XWALK-3115] Update usecase remove xwalk -c test case

linshen:
Impacted TCs num(approved):New 0,Update 0,Delete 1
Test Result: no UT


ibelem:
LGTM

Issue #875, opened 2014-09-11
Refactor scripts of test framework, support resources and web runner


zqzhang:
## Background

In this test repository, we create test cases based on W3C [testharness.js](http://github.com/w3c/testharness.js/) framework, Samsung [unitcommon.js](https://github.com/crosswalk-project/crosswalk-test-suite/blob/master/webapi/tct-alarm-tizen-tests/alarm/support/unitcommon.js), and run the test cases based on Intel [webrunner](https://github.com/testkit/webrunner).

At current being, these scripts are located in each test suite for historical reason. Thus make them lots of copies. See detail analysis [here](http://zqzhang.github.io/blog/2014/07/10/crosswalk-test-suite-analysis.html).

## Proposal

* Create `resources/` in the root of this repository to hold below scripts. Here doesn't use sub-module because we can put all support resources into this directory. List the script files and number of copies here.
  * testharness.css (224)
  * testharness.js (224)
  * testharnessreport.js (223)
  * unitcommon.js (49)
  * jquery.js (6)
  * jquery-1.10.2.min.js (220)
  * jquery.min.js (3)
  * jquery.mobile.js (6)
  * jqueryprogressbar.js (2)
  * slider.tooltip.js (11) see also #866
  * testcase.xsl (197)
  * testresult.xsl (169)
  * tests.css (201)
* Make `webrunner` sub-module for [webrunner](https://github.com/testkit/webrunner) in the root of this repository
  * Please note that this contains `jquery-1.10.2.min.js` too.
  * testrunner.js (220)

To easy the test suite packaging, create a script `tools/build/copy.py` to copy the scripts above to current being placeholder.

zqzhang:
Except several versions of jQuery, these thoughts have been done. Thank you all.

PR #1610, opened 2015-01-08
[webapi][XWALK-3323] Fix improper tct-package-tizen-tests input values

sstolinski:
This patch fixes improper values in conversion table
which are used to validate conversion.

### DIFF
--- a/webapi/tct-package-tizen-tests/package/support/unitcommon.js
+++ b/webapi/tct-package-tizen-tests/package/support/unitcommon.js
@@ -116,17 +116,18 @@ function getTypeConversionExceptions(conversionType, isOptional) {
     switch (conversionType) {
         case "enum":
             conversionTable = [
-                [undefined, exceptionName],
-                [null, exceptionName],
                 [0, exceptionName],
                 [true, exceptionName],
                 ["dummyInvalidEnumValue", exceptionName],
                 [{ }, exceptionName]
             ];
+            if (!isOptional) {
+                conversionTable.push([null, exceptionName]);
+                conversionTable.push([undefined, exceptionName]);
+            }
             break;
         case "double":
             conversionTable = [
-                [undefined, exceptionName],
                 [NaN, exceptionName],
                 [Number.POSITIVE_INFINITY, exceptionName],
                 [Number.NEGATIVE_INFINITY, exceptionName],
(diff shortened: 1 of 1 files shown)


zqzhang:
Literally, I think, this update looks good.

However, I would expect an unique unitcommon.js in all test suite.

I will talk to Samsung Web TCT to see if they want this change. Stay tune.

sstolinski:
@zqzhang I need this patch to fix https://crosswalk-project.org/jira/browse/XWALK-3248.
If it is possible please merge it. If unitcommon.js should be changed/moved/etc. I change it in next commit.

tiwanek:
I'm not sure if this is testsuite fault:
http://www.w3.org/TR/WebIDL/#TreatUndefinedAs
http://www.w3.org/TR/WebIDL/#dfn-optional-argument

tiwanek:
Check: 
https://lists.w3.org/Archives/Public/public-script-coord/2015JanMar/0012.html

Please, rework this to described behaviour (null value case).

zqzhang:
@tiwanek @sstolinski if I read the thread correctly, passing `undefined` as optional nullable param shall not throw. And that is why we need to rework here.

tiwanek:
@zqzhang 
I understand it same way.

@sstolinski 
What about cases "enum" and "double"?

sstolinski:
@tiwanek  Done

tiwanek:
Thanks. For me, it looks good.

zqzhang:
Many thanks.

zqzhang:
@sstolinski feel free to update all `unitcommon.js` files.

BTW, I have been thinking of maintaining only one copy of such file at #875; is that better?

zqzhang:
This has been applied to IVI TCT, thanks.

sstolinski:
I will update all unitcommon.js.

 

IMHO it is good idea.

But some the files are not copies – there are approximately 4-5 different versions (some of them easy to merge).

So at the result 2-3 versions of unitcommon.js will be exists.

 

cid:image001.png@01CE8F77.458ECDF0

 

Sebastian Stoliński

Samsung R&D Institute Poland

Samsung Electronics

 <mailto:s.stolinski@samsung.com> s.stolinski@samsung.com

PR #1613, opened 2015-01-09
[wrt][xwalk-3317] Update Command Line  case for app fullscreen

linshen:
Impacted tests(approved): new 0,update 4,delete 0
Unit test result summary: pass 1,fail 3, block 0

Fail reason:after update test cases by using config.xml viewmodes="fullscreen",it's effective and i can see app launched with fullscreen window,but those test cases contain other test function(eg:Modify the "/usr/lib/systemd/user/xwalk.service" file by 'add --disable-device-capabilities'),they test failed


ibelem:
Hi Can you work with Dalian team to find out the bug ID of "Modify the "/usr/lib/systemd/user/xwalk.service" file by 'add --disable-device-capabilities' ?

Anyway, for fullscreen update task itself, LGTM.

PR #1616, opened 2015-01-12
Update 2 manual cases to auto cases for webstorage

yunxliu:
Unit test platform: [Android]


zqzhang:
@haoxli if we cannot submit a test case for W3C specs to W3C repos, I think, we shall remove the test case rather spend more efforts on test automation. I will talk to you all later about this.

CC @jennycao

PR #1636, opened 2015-01-15
[usecase] Remove test miniLZMACompressionUI from web runtime

yinhaichao:
- To organize crosswalk mini tests better
- Remove test miniLZMACompressionUI from usecase/usecase-wrt-android-tests


ibelem:
LGTM

PR #1637, opened 2015-01-15
[usecase] Update test suite usecase-mini-wrt-android-test name

yinhaichao:
- To organize crosswalk mini test cases better and standard
- Update test suite name from usecase-mini-wrt-android-test to usecase-miniwrt-android-test


ibelem:
LGTM

PR #1638, opened 2015-01-15
Remove the temporary log file

zhuyongyong:
- This covert file is meaningless， need to remove it.


zqzhang:
This has nothing to do with [embedding]. You just mistook submit that within a pull request for embedding api.

PR #1640, opened 2015-01-15
[usecase][xwalk-3355] Add show and hide test for WRT

yinhaichao:
Impacted tests(designed): new 1, update 0, delete 0
Unit test platform: [tizen]
Unit test result summary: pass 0, fail 0, block 1

Block reason:feature triaged
https://crosswalk-project.org/jira/browse/XWALK-3341


ibelem:
@yinhaichao What's the result of step 4? I suggest to remove step 4 directly.

yinhaichao:
@ibelem Your suggestion is right, uninstall operate is not the point of this test and that it had been covered by other test. 
I will remove step 4 directly. 
Thanks for the advice.

ibelem:
LGTM

ibelem:
Let's wait the bug status then update it to 'approved'

PR #1642, opened 2015-01-15
[misc] Update iSmartbook relevant status

yinhaichao:
- Update iSmartbook relevant tests status from approved to designed since they are not in weekly testing


ibelem:
LGTM

PR #1646, opened 2015-01-16
[wrt][XWALK-3355] Update case status with right rules

yinhaichao:
Update case status with following rules:
 - Crosswalk feature in "RESOLVED" status: APPROVED
 - Crosswalk feature in "ACCEPTED" status: DESIGNED
 - Crosswalk test case tasks from bugs: APPROVED with bug ID in UT result


ibelem:
Task: XWALK-3355 Bug: XWALK-3341

PR #1651, opened 2015-01-16
[webapi][XWALK-3363] Add test for CORS DELETE request-type

kangxu:
Unit test platform: [Android]


zqzhang:
Is this test qualified for W3C?

kangxu:
@zqzhang 
  Description on cors DELETE request-type are not detailed enough in W3C spec,  I add this test just for covering this issue [XWALK-3363](https://crosswalk-project.org/jira/browse/XWALK-3363). If submitted to w3c, checkpoints are not strict enough and I extractted test point from spec example mainly.

zqzhang:
Better submit to W3C; if not, we don't need such test either.

haoxli:
I check the CORS spec(http://www.w3.org/TR/cors/), i think we don't need the test for DELETE request-type. close it.

kangxu:
No detailed info found in spec, so close PR.

PR #1652, opened 2015-01-16
[webapi] Fix improper conversionTable values

sstolinski:
This patch fixes improper values in conversion table
which are used to validate conversion.


sstolinski:
@zqzhang Files unitcommon.js have been updated.

PR #1655, opened 2015-01-17
[webapi][XWALK-2955] Update reftests which can not automation to manual

chenxix:
- These tests can not automation,just change the test type.

Impacted tests(approved):new 0,update 11,delete 0 
Unit test platform:[TizenIVI] 
Unit test result summary:pass 11,fail 0,block 0


zqzhang:
Proper indent commit message, please.

Please explain why these tests cannot be automated?

chenxix:
As talked at the meeting just before, from a visual point of view these tests are all passed, but  are all failed when test by testkit-lite tool. These tests enlarge a 1x1-green.png to cover a 100x100 div element. the render result is not the same as a pure 100x100 green div through image comparison tool.  We can not build the same render result by another way.

zqzhang:
OK; to update the commit message, please using 'git commit -a --amend', not in the github UI. Please update them all for your PRs.

Another, if only one item, no need '-' in commit message.

PR #1665, opened 2015-01-17
[webapi][XWALK-3337] Update subcase num for SIMD

JianfengXu:
- Case number isn't consistent in tests.xml and tests.full.xml, so update.


zqzhang:
@haoxli seems we need to clean the xmlsimplifier.py and publish it.

PR #1672, opened 2015-01-19
[cordova] Remove unnecessary configure files

yunxiaoxie:
- For android only, remove "config.xml,inst.wgt.py,inst.xpk.py, manifest.json" and update suite.json to keep android configure only.


Honry:
Please also remove the mainfest.json file, seems it is useless for android only suite. Then verify the package build.

yunxiaoxie:
@Honry updated, remove manifest.json and update suite.json.

yunxiaoxie:
@Honry all cordova packages can be build and run successfully when removed manifest.json.

PR #1676, opened 2015-01-19
Fix unitcommon for application tests

tiwanek:
Passing explicitly undefined to function should not throw.
Therefore test should accept passing undefined.

Follow:
https://lists.w3.org/Archives/Public/public-script-coord/2015JanMar/0012.html

This is same case as in PR:
https://github.com/crosswalk-project/crosswalk-test-suite/pull/1610

BUG=XWALK-3395


tiwanek:
Already applied.

PR #1696, opened 2015-01-21
[webapi][XWALK-3415] Update CSP tests execution type in tests.full.xml

JianfengXu:
- Auto case number isn't consistent in tests.xml and tests.full.xml, so update tests.full.xml


zqzhang:
How about RSA code?

zqzhang:
@haoxli this is also a requirement on the xmlsimplifier.py; with that tool; such kind of issues can be avoided.

JianfengXu:
RSA does not support ref tests, so they are already changed to manual tests.

PR #1697, opened 2015-01-21
[usecase-wrt] Update bootstrap framework to usecase-wrt-android

yunxliu:
- Update usecase-wrt-android samples for bootstrap framework


jennycao:
@yunxliu Pls solve the conflict

PR #1720, opened 2015-01-23
Apdat changes on the new API onXWalkReady()

halton:
Crosswalk environment is not available under two scenarios:
* Shared mode: Need download from app store like Google Play
* CrosswalkMini: The libxwalkcore.so is compressed with lzma, need to
  decompress it first.

org.xwalk.core.XWalkActivity.onXWalkReady() is introduced by for the
purpose of do XWalkView related things after Crosswalk environment is
available to use . Refer to PR
https://github.com/crosswalk-project/crosswalk/pull/2780

Changes need to be adapt this new API:
* XWalkView initialized code moved from onCreate() to onXWalkReady().
* Activity should inheriate from org.xwalk.core.XWalkActivity
instead of android.app.Activity.
* In AndroidManifest.xml, android:name should change to
  org.xwalk.core.XWalkApplication.

XWalkBaseActivity.java is not needed anymore because compiler will
report onXWalkReady() need to be implemented. We move the deceleration
of mXWalkView into activities where this var is needed.


halton:
@jennycao @Belem please reivew. I do not know TIna's id, if needed please help to include her.

jennycao:
Thanks for the PR

PR #1727, opened 2015-01-26
[wrt][xwalk-3442] Add negative  cases for  required version

yinhaichao:
- Test that the web runtime can not install the app when the required version Equal to current tizen system version but the tizen system version is higher than current tizen web api version
- Test that the web runtime can not install the app when the required version equal to tizen web api version but not a float value (eg: required_version=2)
- Test that the web runtime can not install the app when the required version equal to tizen web api version but not a float value (eg: required_version=2.2.0)
- Test that the web runtime can not install the app when the required version is not set
- Test that the web runtime can not install the app when the required version is empty
- Fail reason:xwalk-3483

Impacted tests(approved): new 5, update 0, delete 0
Unit test platform: [Tizen]
Unit test result summary: pass 3, fail 2, block 0


ibelem:
LGTM

PR #1728, opened 2015-01-26
[webapi] Fix of widget packaging' test (as.wgt)

rynkowsg:
as.wgt enumerates whitespace characters.
That list was incorrect, because the test consider x180E.
x180E is not a whitespace starting from The Unicode Standard 6.3.0.
I removed that char from test.

BUG=XWALK-2982


zqzhang:
Good catch. 0x180E is now Mongolian vowel separator.

PR #1750, opened 2015-01-29
[usecase] Update replace some of keyword for wrt

yinhaichao:
- Update "mini" to "lite"
- Update "miniwrt" to "litewrt"
- Update other related file about "mini"


ibelem:
LGTM

PR #1751, opened 2015-01-29
[wrt] Update test.full.xml file format error

yinhaichao:
- Update test.full.xml file format error in tct-ext02-wrt-tests


ibelem:
LGTM

PR #1753, opened 2015-01-29
[webapi] Sync reference files for CSS Color tests to latest W3C

chenxix:
Impacted tests(approved):new 0,update 4,delete 0
Unit test platform:[android]
Unit test result summary:pass 1,fail 3,block 0


zqzhang:
@chenxix I would expect the commit message says something like "Sync reference files for CSS Color tests to latest W3C", or "Integrate reference files for CSS Color tests to latest W3C"; so that people knows you have done in upstream.

zqzhang:
Better with the upstream PR link(s) in the commit message.

PR #1758, opened 2015-01-30
[webapi][XWALK-3459] Update common script for NFC API

kangxu:
- Add checking for testharness to avoid outer script duplicate

Impacted tests(approved): new 0, update 58, delete 0
Unit test platform: [IVI]
Unit test result summary: pass 58, fail 0, block 0

### DIFF
--- a/webapi/webapi-nfc-w3c-tests/nfc/support/nfc_common.js
+++ b/webapi/webapi-nfc-w3c-tests/nfc/support/nfc_common.js
@@ -29,7 +29,14 @@ Authors:
 
 */
 
-document.write("<script language=\"javascript\" src=\"..\/resources\/testharness.js\"><\/script>");
-document.write("<script language=\"javascript\" src=\"..\/resources\/testharnessreport.js\"><\/script>");
+var head_src = document.head.innerHTML;
+
+if (head_src.indexOf('testharness.js') === -1) {
+    document.write("<script src=\"..\/resources\/testharness.js\"><\/script>");
+}
+
+if (head_src.indexOf('testharnessreport.js') === -1) {
+    document.write("<script src=\"..\/resources\/testharnessreport.js\"><\/script>");
+}
 
 var nfcManager = navigator.nfc;
--- a/webapi/webapi-nfc-w3c-tests/nfc/support/unitcommon.js
+++ b/webapi/webapi-nfc-w3c-tests/nfc/support/unitcommon.js
@@ -46,13 +46,14 @@ SECURITY_EXCEPTION = {name: 'SecurityError'};
 
 
 (function () {
-   var head_src = document.head.innerHTML;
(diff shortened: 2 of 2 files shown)


webapi/webapi-nfc-w3c-tests/nfc/support/nfc_common.js
```diff
@@ -29,7 +29,12 @@ Authors:
 
 */
 
-document.write("<script language=\"javascript\" src=\"..\/resources\/testharness.js\"><\/script>");
-document.write("<script language=\"javascript\" src=\"..\/resources\/testharnessreport.js\"><\/script>");
+var head_src = document.head.innerHTML;
+if (head_src.search(/\/testharness.js\W/) === -1) {
```
zqzhang: Hmm, if the `nfc_common.js` script is loaded in the head and before `testharness` scripts loaded, this may not work. See http://stackoverflow.com/questions/9849204/javascript-in-html-head-innerhtml-not-working

Example:
```
<head>
  <script src="support/nfc_common.js"></script>
  <script src="../resources/testharness.js"></script>
</head>
```

Another, why not use `String.indexOf()` rather than `String.search()`?

webapi/webapi-nfc-w3c-tests/nfc/support/nfc_common.js
```diff
@@ -29,7 +29,12 @@ Authors:
...
+    document.write('<script language="javascript" src="../resources/testharness.js"></script>\n');
```
zqzhang: The `language` attribute on `script` has been deprecated in HTML5, see http://www.w3.org/TR/html5-diff/

Moreover, though `type="text/javascript"` is the correct syntax, it overrides the default scripting language. We can also safe leave it out. See
http://stackoverflow.com/questions/2626563/leaving-out-type-text-javascript-language-javascript

zqzhang:
Hm, let's avoid `document.write`. Please remove them from the support file and update test files in need.

webapi/webapi-nfc-w3c-tests/nfc/support/nfc_common.js
kangxu: @zqzhang 
- Similar codes exist in both nfc_common.js and unitcommon.js. 
In these scripts, we will add testharness.js/testharnessreport.js scripts to test page if related script not exist. This way can avoid the repeated load of testharness.js/testharnessreport.js. The repeated load can lead to issue XWALK-3459.

kangxu:
@zqzhang 
I have updated it following your comments.
-  Replace String.search() with String.indexOf()
-  Remove language attribute on script

I still keep document.write.  If we update common script with others way like append(),  the load of outer script is asynchronous. It may lead to that testharness.js/testharnessreport.js have not been loaded successfully while test process has begin.

PR #1761, opened 2015-01-30
Deprecate PDF documents

zqzhang:


zqzhang:
@jennycao @ibelem @cicili

PR #1766, opened 2015-01-30
Update app storage path from ".config" to "apps_rw"

yinhaichao:
- Update app storage path from ".config" to "apps_rw"
  in behavior/misc/usecase/webapi/wrt folder


ibelem:
@yinhaichao Any UT results about them?

ibelem:
LGTM

PR #1780, opened 2015-02-02
[tools] Update web.py to repair the bug of clicking button

chenxix:
Add exception handling when button is not clickable.

### DIFF
--- a/tools/atip/atip/web/web.py
+++ b/tools/atip/atip/web/web.py
@@ -357,9 +357,13 @@ class WebAPP(common.APP):
         element = self.__get_element_by_key(key, display)
         print "%s == %s" % (element.get_attribute("id"), element.get_attribute("class"))
         if element:
-            element.click()
-            return True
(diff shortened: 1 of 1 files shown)


Honry:
Please update prefix to "[ATIP]", specific for atip PRs

cicili:
LGTM

tools/atip/atip/web/web.py
```diff
@@ -407,8 +411,13 @@ def click_element_coords(self, x, y, key, display=True):
     def fill_element_by_key(self, key, text, display=True):
         element = self.__get_element_by_key(key, display)
         if element:
-            element.send_keys(text)
-            return True
+            try:
+                element.send_keys(text)
+                return True
+            except WebDriverException :
```
Honry: I think adding workaround method into exception does not make sense, as the program cannot catch the exception if there is something wrong in try section.

PR #1788, opened 2015-02-03
[wrt] Delete inst.sh.xpk and relative file

yinhaichao:
- Delete inst.sh.xpk file in wrt
- Delete inst.sh.xpk relative file in wrt


ibelem:
LGTM

PR #1805, opened 2015-02-04
[usecase-wrt] Delete the duplicate files for usecase-wrt-android

yunxliu:
- These usecases has been moved to 'samples' folder, so delete them from 'tests' folder


ibelem:
LGTM

PR #1809, opened 2015-02-05
[xwalk-3375] Remove inst.sh.*

haoxli:
- We have used inst.py instead of inst.sh
- The *.spec is useless now


zqzhang:
Any objections? @ibelem @cicili @jennycao @yugang

PR #1810, opened 2015-02-05
[wrt][xwalk-3558] Update /home/app code by multi-user in wrt scripts

yinhaichao:
- Update some /home/app code by multi-user in wrt script
- Update two shell script file in webapi/misc

Task: xwalk-3558


ibelem:
LGTM

PR #1815, opened 2015-02-05
[xwalk-3538] Update manual tests for normal user desc

yinhaichao:
- Update manual tests for normal user desc in wrt/usecase
- Optimal normal user description

Task: xwalk-3538


ibelem:
LGTM

PR #1819, opened 2015-02-05
[xwalk-3558] Update /home/app code by multi-user in wrt scripts

yinhaichao:
- Because of can not get the value of 'TIZEN_USER' when
  execute shell script on ivi device, update the script
  by new way in batch at here
- Talk with the testkit-lite developer, set the 'TIZEN_USER'
  environment variable value on ivi device will not be achieved


ibelem:
@yinhaichao please get tizen_user in common only.

ibelem:
LGTM

PR #1822, opened 2015-02-06
[XWALK-3542] Update docs for multiuser support

zqzhang:
Change the app and developer usernames to normaluser.
The documents shall be revised in unity style.


zqzhang:
@ibelem @yugang please take a look at this change.

ibelem:
@zqzhang Suggest adding a line to explain "normaluser".

zqzhang:
Thanks, @ibelem.

@yugang says it is better use `tester` rather than `normaluser`. I will submit another PR.

PR #1823, opened 2015-02-06
[doc] Use tester rather than normaluser

zqzhang:


zqzhang:
@yugang is this OK to you?

PR #1848, opened 2015-02-11
[webapi][XWALK-3392] Replace the existing case from W3C for Transitions

yunxliu:
- Due to the spec update, the test is blocked by not support error thrown:
  window.getComputedStyle(document.getElementById(elementId)).getPropertyCSSValue(property);
  Can not find method to instead, so delete these tests and replace these case from W3C

Impacted tests(approved): new 15, update 0, delete 0
Unit test platform: [Android]
Unit test result summary: pass 15, fail 0, block 0


zqzhang:
> Due to the spec update, the test is blocked by not support error thrown: window.getComputedStyle(document.getElementById(elementId)).getPropertyCSSValue(property);

Could you please point me which API threw error, `getComputedStyle` or `getPropertyCSSValue`?

```
$ grep 'getComputedStyle' -rl * | wc -l
276
$ grep 'getPropertyCSSValue' -rl * | wc -l
2
$ grep 'getPropertyValue' -rl * | wc -l
239
```

yunxliu:
The spec link: http://www.w3.org/TR/cssom/#the-cssstyledeclaration-interface, getPropertyCSSValue has been removed.

PR #1849, opened 2015-02-11
[wrt] Update test.xml format error in tct-ex02-wrt-tests

yinhaichao:
- The test.xml file missing one '</description>' node


ibelem:
LGTM

PR #1853, opened 2015-02-11
[webapi][OTCIS-4450] Update xml for Fonts

yunxiaoxie:
- Due to testkit-lite has been updated to support both stub and webdriver mode to run all tests, so update xml files to suite.

 Unit test result summary: the suite can run well on android.


Honry:
LGTM

PR #1855, opened 2015-02-11
[webapi][OTCIS-4450] Update xml for 4 suites

kangxu:
- Update xml for 3dtransforms, audio, backgrounds and colors

- Due to testkit-lite has been updated to support both stub and webdriver
mode to run all tests, so update xml files to suite

Unit test result summary: suites can run well on android


Honry:
LGTM

PR #1860, opened 2015-02-11
[wrt] Update pyunit TIZEN_USER for default value

yinhaichao:
Impacted tests(approved): new 0, update 264, delete 0
Unit test platform: [Tizen]
Unit test result summary: pass 264, fail 0, block 0


ibelem:
Because our target will be Tizen Common now, so i just suggest setting default user to "guest". It's only the initial value, we should get final value from testkit-lite anyway.

Do you agree?

ibelem:
@yinhaichao Please update the default value to "app" , i don't have strong opinion about it.

ibelem:
LGTM

PR #1862, opened 2015-02-12
[webapi][OTCIS-4450] Update xml for 3 suites

JianfengXu:
- Update xml for svg, text and multicolumn
- Due to testkit-lite has been updated to support both and webdriver mode to run all tests, so update xml files to suit


Honry:
LGTM

PR #1864, opened 2015-02-12
[webapi] save the bdd related html files outside the subapp

chenxix:
The related html files are saved together with the features, it can not share common files with others. so move them to the root of suite.


Honry:
LGTM

PR #1866, opened 2015-02-12
[webapi] save the webstorage bdd html files outside the subapp

chenxix:
- Change the test type from script to js.
- Change the bdd test type from auto to manual.
- The related html files are saved together with the features files, it cannot share common files with others. so move them to the root path of the suite.

Impacted tests(approved):new 0,update 2,delete 0
Unit test platform:[android]
Unit test result summary:pass 2,fail 0,block 0


Honry:
Please submit your UT result, and update the commit info, as this PR is not only unify the source code.

PR #1893, opened 2015-02-28
[atip] Add new steps for searching element by specified attribute

chenxix:
- Add new steps to find element by given attribute
- Modify the related feature files

Impacted tests(approved):new 0,update 4,delete 0
Unit test platform:[Android]
Unit test result summary:pass 4,fail 0,block 0

### DIFF
--- a/tools/atip/atip/web/web.py
+++ b/tools/atip/atip/web/web.py
@@ -452,6 +452,16 @@ class WebAPP(common.APP):
             return True
         return False
 
+# * The method click_element_by_key will fail when VKB shelter the button, and js can avoid this issue.
+    def click_element_by_id_with_js(self, key, display=True):
+        element = self.__get_element_by_key_attr(key, "id", display)
+        print "%s == %s" % (element.get_attribute("id"), element.get_attribute("class"))
+        if element:
+            js_script = 'document.getElementById(\"' + key + '\").click()'
+            self.driver.execute_script(js_script)
+            return True
+        return False
+
     def click_element_coords(self, x, y, key, display=True):
         element = self.__get_element_by_key(key, display)
         if element:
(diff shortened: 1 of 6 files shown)


usecase/usecase-webapi-xwalk-tests/testscripts/steps/steps.py
```diff
@@ -47,6 +47,14 @@ def check_type_by_key(context, key, expecttype):
 def check_type_by_key(context, p_name, key):
     assert context.app.save_content(p_name, key)
 
+@step(u'I commit message by click "{key}"')
```
Honry: This step name elusive, here you want to click element by id with js method. Please update it.

usecase/usecase-webapi-xwalk-tests/testscripts/steps/steps.py
```diff
@@ -47,6 +47,14 @@ def check_type_by_key(context, key, expecttype):
...
+def i_click(context, key):
+    assert context.app.click_element_by_id_with_js(key)
+
+@step(u'I fill in element "{key}" by "{attr}" with "{text}"')
+def fill_with_text(context, key, attr, text):
```
Honry: Please rename this function name, otherwise it will cause repetition when remove to ATIP.

Honry:
LGTM 
@cicili please take a review.

tools/atip/atip/web/web.py
```diff
@@ -452,6 +452,15 @@ def click_element_by_key(self, key, display=True):
             return True
         return False
 
+    def click_element_by_id_with_js(self, key, display=True):
```
cicili: please add description of this method, explain why use js here to do click, but not direct element.js. Just because element.click will fail when VKB shelter the button, and js can avoid this issue. Then user can choose whether to use this method.

PR #1895, opened 2015-02-28
[behavior][XWALK-3167] Add SSO Exception sample.

JianfengXu:
- Add the sample to verify XWALK-3167
- SSO is not supported on tizen-common now, so change the status to designed.


jennycao:
@JianfengXu reject this PR, since The activity on Tizen would be suspend, include test execution on tizen common, new case dev, bug fix.

PR #1933, opened 2015-03-12
[webapi]Update suite.json file in extra and webstorage suite

chenxix:
Add copylist sub section to apk-aio otherwise all-in-one package will lose the related html files for sub app.


zqzhang:
This is a duplicate of #1921

zqzhang:
Oh, this is for the Crosswalk-12 branch. Thanks.

PR #1940, opened 2015-03-13
[tools]Perfect update.py to support cordova4.0 suits packing.

jiajiax:


jiajiax:
I have verified that cordova4.0 suites can be packed successfully with this patch.

Honry:
LGTM

PR #1963, opened 2015-03-20
Added OSX info and updated link to apktool.

truegilb:
After running the tool on OSX, I thought that it will good to update some info for users on the same platform.


ibelem:
LGTM, @truegilb thank you for the updates! :)

PR #1971, opened 2015-03-23
[atip] Move new added steps from suite to atip

chenxix:
- Before the new added steps are saved together with the tests, just move them to atip now.


Honry:
As you move steps to a common place, this may caused duplication, please verify all relevant tests.

tools/atip/atip/web/steps.py
```diff
@@ -179,3 +181,101 @@ def i_accept_alert(context):
...
+def save_div(context, key, pic_name):
+    assert context.app.save_div_as_picture(key, pic_name, width=0, height=0)
+
+
+@step(u'I save div "{key}" as "{pic_name}" with width "{width}" and height "{height}"')
+def save_div(context, key, pic_name, width, height):
+    assert context.app.save_div_as_picture(key, pic_name, width, height)
+
+
+@step(u'pic "{pic1}" and pic "{pic2}" should be more than "{similarity}" similar')
```
cicili: why not put the steps into common folder? as the called function under common.
Same requirement for other steps whose function under common.

chenxix:
Yes, I have check all the bdd tests already, and the result is OK.

cicili:
LGTM

Honry:
LGTM

PR #1973, opened 2015-03-25
[usecase] Add usecase test suite for linux

yinhaichao:
Impacted tests(approved): new 25, update 0, delete 0
Unit test platform: [Linux]
Unit test result summary: pass 20, fail 5, block 0

Fail reason:
    - ApplicationIndexDB_Upgrade case cannot read property 'value' of undefined
    - Cookies cases can not alert display


ibelem:
@yinhaichao , could you please file bugs for failed test cases?

PR #1977, opened 2015-03-25
[tools]Add pack_deb.py file for linux

yinhaichao:
Unit test platform: [Linux]

### DIFF
--- a/tools/build/pack_deb.py
+++ b/tools/build/pack_deb.py
@@ -1,0 +1,772 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2015 Intel Corporation.
+#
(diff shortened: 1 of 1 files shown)


ibelem:
@jennycao Could you please help to check if we need a separate pack_linux.py or merge the pack_linux.py into pack.py?

jennycao:
suggest to keep seperate files, we could merge if there is more request.
another suggestion, it would be better pack_deb.py ? since pack.py also executed on linux , right ?

ibelem:
Thanks @jennycao  for the suggestions! @yinhaichao please update the file name. :)

ibelem:
LGTM

PR #1964, opened 2015-03-20
[webapi] Integrate reference file and test from W3C for CSS3UI

yunxliu:
- Integrate selector-required.html from W3C to instead CSS3UI_pseudo-classes_required-manual.html
- Integrate reference file from W3C

Impacted tests(approved): new 0, update 1, delete 0
Unit test platform: [Android]
Unit test result summary: pass 1, fail 0, block 0

PR #1978, opened 2015-03-25
[webapi] Integrate test from W3C for extra

yunxliu:
- Integrate form-validation-validity-tooLong.html from W3C to instead textarea_element_maxLength-manual.html

Unit test platform: [Android]


zqzhang:
Though this is a BDD automated test now, I want to say again as in https://github.com/crosswalk-project/crosswalk-test-suite/pull/1964#issuecomment-85343953

> For the test cases (for W3C specs) you are trying to automate, please first check whether it is qualified for submitting to W3C. If W3C won't accept the test, we'd better remove it rather than automate it. This is a general rule.

cicili:
@zqzhang , this case is liuyun found an auto case from upstream to replace current manual test, and have no relation with bdd, please help to review this patch

PR #1980, opened 2015-03-26
[misc] Update tests.full/tests.xml file in webappmanu-linux-tests

yinhaichao:
- Update 'status' property from 'designed' to 'approved' in tests.full.xml
- Add new cases list in tests.xml
- Update 'test_script_entry' property value


ibelem:
@yinhaichao Is they any reason to update path from "webappmanu-linux-tests" to "webappmanulinuxtests"?

ibelem:
LGTM.

PR #1993, opened 2015-03-30
[webapi][XWALK-3646] Remove audio and video blok tests

xinliux:
These tests is bloked because of some events failed to invoke, and these check points are not good for w3c,  so remove thest tests.


zqzhang:
Seems you still not fix the issue.

zqzhang:
Hmm, I prefer removing these tests if they are not qualified to submit to W3C.

zqzhang:
Ping @xinliux

xinliux:
I will remove these cases as soon as possible.

PR #2005, opened 2015-03-31
[misc|usecase] Update inst.deb.py script file

yinhaichao:
- Update code to native mode in order that run in deepin platform
- Remove some unwanted function
- Test successfully


ibelem:
LGTM

PR #2009, opened 2015-04-01
[atip] Update atip to adapt if there is no data.conf

chenxix:
- If there is no data.conf file, the test tool will work OK.
- Save the result pictures together with the baseline


chenxix:
I have test the new atip on tct-webstorage-w3c-tests/tct-extra-html5-tests/css3-color-app/css3-backgrounds-app ,  and the result is OK.

PR #2012, opened 2015-04-01
[tools]Update pack.py for packing cordova4.0 suites.

jiajiax:


Honry:
LGTM

jiajiax:
I have checked that cordova4.0 suites can be packed successfully with this patch.

PR #2022, opened 2015-04-07
[apptools]Update apptools test suite for linux

yinhaichao:
Unit test platform: [Linux]
Unit test result summary: pass 17, fail 2, block 0

Update:
* update the run method, and tests xml file.  add new method to prevent process is engaged
* update the build method, add the delete temporary files

Fail reason:
* Create pkgname like "org.example.1234test/org.example.1234" successfully, but expect failed

Buid id: xwalk-3818


Honry:
LGTM

PR #2027, opened 2015-04-07
[wrt]Add auto csp test for linux

yinhaichao:
- There are two wrt test issues for testskit in deepin:
  For type=wrt, it run succussful, but testkit get the
  test result as N/A, even if the test run PASS.

  In my deepin. every tests will be launched twice,
  but the test num in result xml is correct.

Impacted tests(approved): new 5, update, 0, delete 0
Unit test platform: [Linux]
Unit test result summary: pass 0, fail 0, block 5


ibelem:
@yinhaichao @haoxli CSP auto test cases should be in tct-csp-w3c-tests which type='js' i think? http://otcqa.sh.intel.com/wiki/Crosswalk_on_Linux_Desktop_Test_Plan#Crosswalk_on_Debian_and_other_Linux_Distributions 

For any test suites leveraged from web api, keep the web api test suites.

ibelem:
@yinhaichao @haoxli Please double check if the type='wrt' here is correct, if you are using testharness.js, it should be 'js'.

PR #2053, opened 2015-04-13
[tools]Update pack.sh for cordova all-in-one suite shared mode packing.

jiajiax:


Honry:
Please provide the UT result.

jiajiax:
Checked done, It is ok with this patch.

Honry:
LGTM

PR #2061, opened 2015-04-14
[usecase] Update README for the WRT test suite for Linux

zqzhang:
Describe the purpose of the test suite and add new test developer.


zqzhang:
CC @kangxu

PR #2134, opened 2015-04-28
[webapi] Remove unreferenced shell scripts

zqzhang:
Samsung engineer finds these.


zqzhang:
@jiajiax do you use these scripts?

jiajiax:
@zqzhang No, I do not use them.

PR #2153, opened 2015-05-05
pack support for stability testcase

jack-yong:
[XWALK-4096]pack support build test app in java build & build test suite app in js build for stability testcase


zhuyongyong:
LGTM

PR #2154, opened 2015-05-05
[embeddingapi][XWALK-3927]add use case to check XWalkActivity API

jack-yong:
-XWalkActivity is a new API of the Embedding API. So just inherit from it like XWalkViewShellActivity, you can use XWalkView.

Unit test platform: [Android]


jack-yong:
Now XWalkActivity has been used in embeddinapi usecase. so just abort this one.

PR #2155, opened 2015-05-05
[usecase] Fix parser error of wrt-linux tests.full.xml

zqzhang:


zqzhang:
@kangxu please take care of this. Such kind of errors must be avoided in future.

PR #2156, opened 2015-05-05
Add Chromium Feature Verification Guid

chunywang:


zqzhang:
Merge as-is for further improvement.

PR #2192, opened 2015-05-12
[usecase-webapi] Resolve unstable issue in ResourceTiming.feature

haoxli:
When the device performance or network in poor, there is no enough
 time for loading resource, this make the test result unstable.
Add waiting time for 5 second.

Unit test platform: Crosswalk Project for Android 14.43.342.0

BUG=https://crosswalk-project.org/jira/browse/XWALK-4095

### DIFF
--- a/usecase/usecase-webapi-xwalk-tests/testscripts/ResourceTiming.feature
+++ b/usecase/usecase-webapi-xwalk-tests/testscripts/ResourceTiming.feature
@@ -1,8 +1,9 @@
 Feature: Usecase WebAPI
  Scenario: Performance & Optimization/ResourceTiming Test
     When launch "usecase-webapi-xwalk-tests"
-     And I go to "/samples/ResourceTiming/index.html"
-     And I press "requestShow"
+    And I go to "/samples/ResourceTiming/index.html"
+    And I press "requestShow"
+    And I wait for 5 seconds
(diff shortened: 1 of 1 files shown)


zqzhang:
I don't like the magic timeout. Any confidence of the 5 seconds?  Perhaps there is another device needs 15 seconds.

Honry:
This method would be more appropriate: `I should see "{text}" in {timeout:d} seconds`
It would always check the text until time's up, and stop checking when text is found.
Thus we can use an even longer timeout.

haoxli:
@zqzhang if we don't use magic number, it need the runner support asynchronous method.  now we only can set timeout for waiting action. And the time is decided by our test devices, and make it double.

haoxli:
@Honry `I should see "{text}" in {timeout:d} seconds` is not a good choice, we only need waiting at the beginning, not each step.

Honry:
I guess you misunderstand my word, we don't need waiting if the text can be found.

haoxli:
But we also need check the `{text}` in `{area}`, so the exiting step cannot meet the request

Honry:
All right, suggest to implement it in the future, no more comment for current solution. @cicili, any comment or concern?

Honry:
As @cicili takes AL today, I would like to merge it.

PR #2201, opened 2015-05-13
[embeddingapi][XWALK-4052]add usecase to clear the cache

jack-yong:
-Add usecase to clear the cache and disk cache

BUG=https://crosswalk-project.org/jira/browse/XWALK-4052

Unit test platform: [Android]


jennycao:
in XWALK-4052, there are 2 secarios
1. from onPageLoadStop()
2. from a button click call back.()

Current usecase cover the first scenario, Pls also cover the second one in the usecase
:)

PR #2205, opened 2015-05-13
[webapi] Remove invalid check points in HTML5 Audio&Video

haoxli:
The no argument is not belong to invalid tests.
An HTML method’s parameter is an output of http://heycam.github.io/webidl/,
before it checks the real parameter, type, it shall check the IDL JS binding.
Then, TypeError is expected, and this should be checked in IDL tests.

BUG=https://crosswalk-project.org/jira/browse/XWALK-3106
BUG=https://crosswalk-project.org/jira/browse/XWALK-3107


webapi/tct-video-html5-tests/video/video_canPlayType_no_argument.html
```diff
@@ -1,53 +0,0 @@
...
-    <script src="../resources/testharnessreport.js"></script>
-    <script src="support/mediaelement.js"></script>
-  </head>
-  <body>
-    <div id="log"></div>
-    <video id="m">
-        Your browser does not support media elements.
-    </video>
-    <script>
-        canPlayType_no_argument();
```
zqzhang: Please also remove the `canPlayType_no_argument()` from `support/mediaelement.js`.

PR #2229, opened 2015-05-19
[webapi] Correct component of HTML5 Application caches

zqzhang:
Historically the components are from the Tizen API reference
hierarchical structure. The appcache shall be categorized in Storage,
see
https://developer.tizen.org/dev-guide/2.3.0/org.tizen.web.apireference/html/w3c_api/w3c_api_m.html#storage

https://crosswalk-project.org/jira/browse/XWALK-1355


zqzhang:
@jennycao @haoxli @chunywang

PR #2232, opened 2015-05-19
[webapi] Leverage Crosswalk Web API category

zqzhang:
As Kenneth pointed out, App URI is still being used by us and Firefox
OS, so that should stay around.

Though the App URI spec was published by W3C System Application WG, we
leverage Crosswalk Web API category as testing component. See
https://crosswalk-project.org/documentation/apis/web_apis.html

https://crosswalk-project.org/jira/browse/XWALK-388


zqzhang:
@tinazhao @jennycao @haoxli @chunywang

PR #2257, opened 2015-05-26
Format Python scripts using autopep8

zqzhang:
find . -name '*.py' -type f | autopep8 --in-place --aggressive

These scripts shall be checked by pylint one by one.

### DIFF
--- a/apptools/apptools-android-tests/apptools/allpairs.py
+++ b/apptools/apptools-android-tests/apptools/allpairs.py
@@ -34,10 +34,27 @@ import commands
 import shutil
 import comm
 
+
 def generate_cmd():
     comm.setUp()
-    positive_data = ['org.xwalk.tests', 'org.xwalk.t1234', 'org.example.xwal_', 'org.example.te_st', 'or_g.example.xwalk', 'org000.example.xwalk', 'org.example123.xwalk']
-    negative_data = ['org.xwalk', 'test', 'org.example.1234test', 'org.example.1234', '123org.example.xwalk', 'org.123example.xwalk', 'org.example._xwalk', 'org.xwalk.Tests', '_org.example.xwalk']
+    positive_data = [
+        'org.xwalk.tests',
+        'org.xwalk.t1234',
+        'org.example.xwal_',
+        'org.example.te_st',
+        'or_g.example.xwalk',
+        'org000.example.xwalk',
+        'org.example123.xwalk']
+    negative_data = [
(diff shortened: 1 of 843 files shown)


zqzhang:
@ibelem @Honry @jennycao @cicili @haoxli @jiajiax please take a look at these changes.

I would like to hear from @haoxli @jiajiax @zhuyongyong for your verification. Many thanks.

zqzhang:
My Python version:
```
$ python -V
Python 2.7.3
```

haoxli:
I think it have a effect for pyuint tests, there are many cases exists in apptools, wrt, bdd, we can check them first. The pack.py and inst.*.py are same in each test suite, only need verify 2-3 suites.
I will assign to their ownner to check that.

ibelem:
For the diff like below:
  except Exception,e:
    print Exception,"Generate pkgName.py error:",e
  except Exception as e:
    print Exception, "Generate pkgName.py error:", e

 In Python 3.x, using as is required.
 In Python 2.6+, use the as syntax, since it is far less ambiguous and forward compatible with Python 3.x.
 In Python 2.5 and earlier, use the comma version, since as isn't supported.

The requirements for Crosswalk test suite is "Python 2.7 or greater", the change from “,” to “as” works on Python 2.7.6, so I think it’s fine to update the code to “as” which has better compatibilities for Python 2.x and 3.x.

kangxu:
I have tried several suites with these updates,  including:
- usecase-wrt-linux-tests
- webappmanu-linux-tests
- wrt-rt-linux-tests
- tct-notification-w3c-tests

Verification for these suites is ok. Test app can be generated, installed, launched and uninstalled normally. And test case can also be run normally.

yunxliu:
I have tried several suites with these updates, including:
- apptools-android-tests
- apptools-linux-tests
Verification for these suites is ok. Test app can be generated, installed, launched and uninstalled normally. And test case can also be run normally.

chunywang:
I have tested suites with these updates, including:
  tct-animations-css3-tests
  webapi-contactsmanager-w3c-tests
Verification for these suites is ok. Test apps can be generated, installed, launched normally. And test case can also be run normally

JianfengXu:
I have tested tct-colors-css3-tests  with these updates, including:auto tests, manual tests and bdd tests.

Verification for this suite is ok. Test app can be generated, installed, launched normally. And test cases can also be run normally

PR #2289, opened 2015-06-03
[Cordova] Auto the CIRC build

zhuyongyong:
- It's successful to build CIRC apk

BUG=https://crosswalk-project.org/jira/browse/XWALK-4329

### DIFF
--- a/tools/build/pack_cordova_sample.py
+++ b/tools/build/pack_cordova_sample.py
@@ -28,6 +28,7 @@
 # Authors:
 #         Fan, Yugang <yugang.fan@intel.com>
 #         Lin, Wanming <wanming.lin@intel.com>
+#         Zhu, Yongyong <yongyongx.zhu@intel.com>
 
 import os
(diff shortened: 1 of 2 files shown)


tools/build/pack_cordova_sample.py
```diff
@@ -52,7 +52,7 @@
 TOOL_VERSION = "v0.1"
 VERSION_FILE = "VERSION"
 DEFAULT_CMD_TIMEOUT = 600
-PKG_NAMES = ["gallery", "helloworld", "remotedebugging", "mobilespec"]
+PKG_NAMES = ["gallery", "helloworld", "remotedebugging", "mobilespec", "cordovacirc"]
```
Honry: 'circ' would be better, what do you think?

Honry:
Please drop a mail to @jiajiax about this new requirement, also tell him the precondition.

Honry:
LGTM

PR #2292, opened 2015-06-04
[webapi] Update qunit, khronos harnesses.

BruceDai:
Support running those webapi cases of tct-webgl-nonw3c-tests,
webapi-simd-nonw3c-tests, and webapi-webcl-nonw3c-tests automatically
on Web Testing Service.


JianfengXu:
Both of qunit and khronos harnesses  have  subcase concept, This update is for intergating subcases test results, and there is no impact on existing testing methods. So I think we can accept this update.
It's better describe the update information in  COPYING file.

zqzhang:
@BruceDai please update related COPYING files for the modification accordingly, thanks.

PR #2298, opened 2015-06-05
[usecase-cordova] Update thirdparty plugin path

haoxli:


haoxli:
Verified PASS in usecase cordova 3.6 and 4.0.

Honry:
LGTM

PR #2301, opened 2015-06-05
[webapi] Add pre-condition in README for webspeech

haoxli:
According to XWALK-4232, test device should make sure
that google voice search service is installed.
Add this to README.

BUG=https://crosswalk-project.org/jira/browse/XWALK-4338


webapi/webapi-webspeech-w3c-tests/README
```diff
@@ -7,6 +7,11 @@ This test suite is for testing web speech specification:
 
 * Shentu,Jiazhen <jiazhenx.shentu@intel.com>
 
+## Pre-condition
+
+* Make sure the google voice search service is installed.
```
zqzhang: How to make sure?

PR #2302, opened 2015-06-06
[Crosswalk-14][webapi] Remove invalid check points in HTML5 Audio&Video

haoxli:
The no argument is not belong to invalid tests.
An HTML method’s parameter is an output of http://heycam.github.io/webidl/,
before it checks the real parameter, type, it shall check the IDL JS binding.
Then, TypeError is expected, and this should be checked in IDL tests.

https://crosswalk-project.org/jira/browse/XWALK-3106
https://crosswalk-project.org/jira/browse/XWALK-3107


zqzhang:
idlharness.js tests have covered this checkpoint.

Issue #1680, opened 2015-01-20
Enable CI Status on Pull Requests


zqzhang:
https://github.com/w3c/web-platform-tests has enabled https://travis-ci.org/w3c/web-platform-tests to automatically check every pull request. This is a good practice. Let's enable this continuous integration service here.

Some references:
- https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.md#ci-status-on-pull-requests
- https://github.com/blog/1227-commit-status-api
- https://developer.github.com/v3/repos/statuses/

zqzhang:
@haoxli

haoxli:
@kangxu  Please spend some effort on this recently.

zqzhang:
Because we cannot access to the setting of this repository, let's stop the task.

kangxu:
Step1:
![image](https://cloud.githubusercontent.com/assets/7007826/11290452/c977720a-8f71-11e5-99ea-84c8b30c2e34.png)

kangxu:
Step2:
![image](https://cloud.githubusercontent.com/assets/7007826/11290464/e0234768-8f71-11e5-8a5a-225e85ce5a6a.png)

kangxu:
Step3:
![image](https://cloud.githubusercontent.com/assets/7007826/11290513/7b753fa0-8f72-11e5-9f62-9ce8f9b2c5c8.png)

kangxu:
Step4:
![image](https://cloud.githubusercontent.com/assets/7007826/11290534/b207bf8e-8f72-11e5-8dc8-1ef767aa806c.png)

kangxu:
Step5:
![image](https://cloud.githubusercontent.com/assets/7007826/11290596/308a0448-8f73-11e5-9f4f-8ce33c75a5fd.png)

kangxu:
Step6:
![image](https://cloud.githubusercontent.com/assets/7007826/11290611/4c10f5a0-8f73-11e5-8d80-413ec3cf36af.png)

kangxu:
Step7:
![image](https://cloud.githubusercontent.com/assets/7007826/11290624/6b7848e4-8f73-11e5-8667-197b6343bb3a.png)

kangxu:
Step8:
![image](https://cloud.githubusercontent.com/assets/7007826/11290654/bc753892-8f73-11e5-9e0f-41cbc35427dd.png)

kangxu:
Step9:
[...]

PR #2305, opened 2015-06-08
[tools] Enable CI status pull requests

kangxu:


.travis.yml
```diff
@@ -0,0 +1,4 @@
+language: python
+python:
+  - "2.7"
+script: ./tools/travis/lint.py
```
zqzhang: Please put the script in `tools/format`

tools/travis/lint.py
```diff
@@ -0,0 +1,92 @@
...
+# Authors:
+#         Xu, Kang <kangx.xu@intel.com>
+
+import sys
+import os
+import re
+import os.path
+import commands
+
+Lint_Root = os.path.dirname(os.path.abspath(__file__))
```
zqzhang: Why not use https://github.com/w3c/wpt-tools/blob/master/lint/lint.py with some modification against testharness.js checking?

kangxu:
@zqzhang 
I will update script refer to https://github.com/w3c/wpt-tools/blob/master/lint/lint.py

zqzhang:
@kangxu thank you very much for trying this to fix #1680.

Because we cannot access to the setting of this repository, let's stop the task.

PR #2311, opened 2015-06-10
[tools]Update pack.py to support apk-shared and googleplayer mode in usecase-wrt.

jiajiax:


jennycao:
@wanghongjuan Pls help to review and verify this change

PR #2312, opened 2015-06-10
[apptools-android] Check if MainActivity.java changed for apptools android

yunxliu:
Add 3 test cases to test MainActivity.java changed. The status of these test cases are designed.

https://crosswalk-project.org/jira/browse/XWALK-4114


Honry:
LGTM

PR #2313, opened 2015-06-10
[usecase-webapi] Remove useless css file for usecase webapi

yunxliu:
bootstrap-theme.css is only quoted in bootstrap-theme.css.map, so delete it. Other files aren't quoted, so delete them.


zqzhang:
I prefer `bootstrap.min.css` to `bootstrap.css`.

yunxliu:
bootstrap.css is easy to change style. If bootstrap.min.css is refered, it will hard to modify style.So I think it is better to refer bootstrap.css.

jennycao:
LGTM

PR #2348, opened 2015-06-17
[stability-embeddingapi] Improve the interface of embeddingapi in stability

zhuyongyong:
- Improve the pages show
- Add separate activity for navigation and swype
- Improve the views count to avoid the influence of navigation
- Call mAddViewsButton.performClick() in a accurate time
- Add shared mode support
- Use xwalkview instead of surfaceview and remove textureview
- Something error when create 5 xwalkviews which load different pages, refer to XWALK-4514

BUG=https://crosswalk-project.org/jira/browse/XWALK-4421


haoxli:
@zhuyongyong please add UT

zhuyongyong:
I've regression UT  for  testapp.

jennycao:
merge the large PR first. if any issue. fix with new PR.

PR #2360, opened 2015-06-18
[apptools-android] Add case to test project can be reset for apptools android

yunxliu:
Add 2 cases to test the project can be reset to initial state with clean command. The status of these test cases are designed.

https://crosswalk-project.org/jira/browse/XWALK-4153

### DIFF
--- a/apptools/apptools-android-tests/apptools/clean.py
+++ b/apptools/apptools-android-tests/apptools/clean.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2015 Intel Corporation.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of works must retain the original copyright notice, this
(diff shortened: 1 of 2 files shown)


apptools/apptools-android-tests/apptools/clean.py
```diff
@@ -0,0 +1,72 @@
...
+import commands
+
+
+class TestCrosswalkApptoolsFunctions(unittest.TestCase):
+
+    def test_clean_file(self):
+        comm.setUp()
+        comm.create(self)
+        os.chdir('org.xwalk.test/prj/android/src/org/xwalk/test')
+        fp = open("MainActivity.java", 'r+')
```
Honry: You only check the reset of MainActivity.java, what about other intermediary files?

yunxliu:
I will update this pr when this new feature has implement, I still don't know other intermediary files will be reset to initial state with clean command.

Honry:
Close it because the feature is "Won't Fix".

PR #2370, opened 2015-06-19
[usecase] Delete WebCL usecase

JianfengXu:
Due to WebCL sample has existed in demo-express, so delete it in usecase.


zqzhang:
@jennycao shall we accept this?

jennycao:
yes,since we add it to Demo-express, need to remove the duplicated one in usecase;)

PR #2371, opened 2015-06-19
[cordova] Improve pack_cordova_sample.py 

zhuyongyong:
-When not use "--notclean", it can pack successfully now

BUG=https://crosswalk-project.org/jira/browse/XWALK-4429


Honry:
LGTM

PR #2374, opened 2015-06-19
[usecase] Update feature for SIMD

JianfengXu:
When SIMD button is disable, the FPS value without SIMD will get two times. Due to two FPS values have a very samll differnence and the second value always greater than the first value very little. We will get pass result.

This update is to avoid this problem.

Impacted TCs num(approved): New 0, Update 1, Delete 0
Unit test Platform: Crosswalk Project for Android 15.44.364
Unit test result summary: Pass 0, Fail 1, Blocked 0

Fail analysis: SIMD does not work after rebase to M44. More details in https://crosswalk-project.org/jira/browse/XWALK-4207

### DIFF
--- a/tools/atip/atip/web/steps.py
+++ b/tools/atip/atip/web/steps.py
@@ -135,7 +135,7 @@ def i_press(context, key):
 
 
 @step(u'press "{key_c}" in "{key_p}"')
-def i_press(context, key_p, key_c):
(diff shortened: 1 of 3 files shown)


zqzhang:
@jennycao @ibelem who is taking care of ATIP tests now?

zqzhang:
@Honry PTAL

Honry:
LGTM

PR #2378, opened 2015-06-23
[tools] Sync license of testharness.js to latest

zqzhang:
https://github.com/w3c/testharness.js/pull/133
made an explicit LICENSE of the testharness.js project,
so integrate it here directly, with the integration commit
into README.md as note.

https://crosswalk-project.org/jira/browse/XWALK-4464


jennycao:
LGTM

PR #2379, opened 2015-06-23
[webapi] Integrate upstream tests for Colors

JianfengXu:
Due to the scroll bars are different between source page and ref page, the ref case test failed.
So integrate new upstream tests.

Impacted TCs num(approved): New 15, Update 0, Delete 1
Unit test Platform: Crosswalk Project for Android 15.44.364
Unit test result summary: Pass 15, Fail 0, Blocked 0

BUG=https://crosswalk-project.org/jira/browse/XWALK-4034


zqzhang:
> Due to the scroll bars are different between source page and ref page, the ref case test failed.
> So hide the scroll bar.

No, please don't do such workaround. Please fix the test/ref in W3C repo directly. If you have any question, feel free to ask help in the public-css-testsuite@w3.org mailing list.

zqzhang:
> When the page is loaded, the scroll bar will gradually disappear. So sometimes the screenshot has scroll bar, and sometimes not. A few pages have the same issue.

Is this a feature of Chrome or Crosswalk? Or is it a bug?

Removing the CSS styles, we may figure out a scroll bar bug, right?

Another, I think the test and reference files are not so good, because they make scroll bar, which need user interaction to finish the testing. If a test needs user interaction, it cannot be automated by reftest. So, shall we rewrite (or split) the test/ref files?

zqzhang:
@JianfengXu please sync this with https://github.com/w3c/csswg-test/pull/787

webapi/tct-colors-css3-tests/colors/csswg/t424-hsl-values-b-1.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Test: hsl() values</title>
+<link rel="author" title="L. David Baron" href="http://dbaron.org/" />
+<link rel="author" title="Mozilla Corporation" href="http://mozilla.com/" />
+<link rel="author" title="Intel" href="http://www.intel.com">
+<link rel="help" href="http://www.w3.org/TR/css3-color/#hsl-color" />
+<link rel="match" href="t424-hsl-values-b-1-ref.html" />
```
zqzhang: Seems you put all reference files into reference/ sub-directory, but doesn't update the link here.

Moreover, I think the reference/ sub-directory is not necessary.

PR #2384, opened 2015-06-24
Add ref tests for 3dtransforms

chunywang:
Fail analysis: Actual pic is different with expected pic when crosswalk
render pics. For example: css-transform-3d-rotate3d-X-negative.html,
there is a red border outside black border, but the ref has not.

Impacted tests(approved): new 0, update 40, delete 0
Unit test platform: Crosswalk Project for Android 15.44.361.0
Unit test result summary: pass 38, fail 29, block 0


zqzhang:
Please double whether we can improve these tests to avoid fractional pixel as we talked.

PR #2393, opened 2015-06-25
[apptools-ios] Change execution_type from 'manual' to 'auto' for apptools ios

yunxliu:
Change execution_type from 'manual' to 'auto', since ios has already supported testkit-lite


Honry:
LGTM

PR #2398, opened 2015-06-26
[webapi] Remove duplicate test for css3 colors

kangxu:


zqzhang:
Good catch, any bug report?

kangxu:
@zqzhang No bug report, the entry of removed test is same with test id="css3_colors_tests_entry24"

PR #2402, opened 2015-06-26
[usecase-wrt] Add test for CSP to usecase

wanghongjuan:
Unit test Platform: Android
Android test result summary: Pass 1, Fail 0, Block 0


ibelem:
where is the samples/CSP/res ?

PR #2406, opened 2015-06-26
[embeddingapi] Improve maven build and gradle build for embeddingapi pack

zhuyongyong:
- It's successful to build embeddingapi pkg using maven build and gradle build

https://crosswalk-project.org/jira/browse/XWALK-3784

### DIFF
--- a/embeddingapi/embedding-api-android-tests/README
+++ b/embeddingapi/embedding-api-android-tests/README
@@ -4,7 +4,10 @@ This test suite is for the unit tests of Embedding API crosswalk module with XWa
 
 ## Pre-conditions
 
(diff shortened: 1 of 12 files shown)


jennycao:
@haoyunfeix @qiuzhong 
Pls notice change,
After this PR merged, we will deploy the maven/gradle build into regular(nightly) testsuite release.
Nightly build script may need to update to support it .

qiuzhong:
@jennycao , this PR will cause the same name of package of embeddingapi tests no matter what kind of build way. That means if we first use maven and then use gradle way to build, there will be only a package is generated because the latter will overwrite the former. To fix this issue, we need commit another PR.

PR #2420, opened 2015-07-01
[tools] remove the WTS build script SIMD "-manual" option

qiuzhong:
Before, we add the "-manual" in the test html files for SIMD specially.
Now, SIMD qunit framework is supported to auto execution, so we need
to remove "-manual" for these html files.

https://crosswalk-project.org/jira/browse/XWALK-4509


jennycao:
LGTM

PR #2443, opened 2015-07-06
[wrt] Update test script for packertool and manifest

wanghongjuan:
- Update the naming rule for manifest script
- Update the tests.py for packertool
- The test suites can pack and run normally


ibelem:
LGTM

PR #2456, opened 2015-07-08
[ATIP] Fix improper path of baseline

Honry:
Add test url in webdriver env to redirect baseline path
  BUG=WEBQA-248

### DIFF
--- a/misc/sampleapp-android-tests/testscripts/environment.py
+++ b/misc/sampleapp-android-tests/testscripts/environment.py
@@ -64,6 +64,7 @@ def load_default_config():
         platform.update({"comm-mode": comm_mode})
         platform.update({"device": device})
         webdriver_json.update({"platform": platform})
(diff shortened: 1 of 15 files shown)


zqzhang:
> [ATIP]Fix webqa-248

First please put a space after prefix `[...]`. Because the prefix is used to distinguish test type (or sub-directory), however, here you don't need `[ATIP]`. Better without prefix because this PR touches several sub-directories.

Second, `webqa-248` is something internal; please don't use it; or just as `BUG=WEBQA-248` for a record. Instead, a short message like `Fix <short description of the real issue>` is recommended.

Honry:
@zqzhang, thanks for the comment, update the summary, we appointed add ```[ATIP]``` prefix for ATIP matter, and the ```environment.py``` is part of ATIP, and they are all the same.

zqzhang:
LGTM

PR #2462, opened 2015-07-09
[TC release]Add sampleApp_pack.sh for better maintenance

qiuzhong:
Add sampleApp_pack.sh to crosswalk-test-suite/misc/sampleapp-android-tests
for better maintenance. It's intended to release all the sample apps apk files
while it should not be packed into the test packages. That's why it's in the
blacklist in suite.json file.

https://crosswalk-project.org/jira/browse/XWALK-4555


jennycao:
Thanks;)

PR #2464, opened 2015-07-10
[TC release]Add maven/gradle build way suffix for embeddingapi test s…

qiuzhong:
…uites

We need all maven/gradle/ant build way for all the embeddingapi test suites.
However, each way has the same zip file name after packing. That will cause
the latter build overwrite the former and only produce one zip file. We want
3 zip files in all for each embeddingapi test suite.
To fix this, we add a build way suffix for the zip file name.

https://crosswalk-project.org/jira/browse/XWALK-4528


qiuzhong:
Please notice, adding maven/gradle/ant way on Crosswalk-14 and master branch has different way. Please don't merge this PR to master branch at present. I will submit another PR for master branch.

jennycao:
LGTM

PR #2481, opened 2015-07-14
[tools] Remove Tizen related test suites from WTS

zqzhang:
Web Testing Service (WTS) shall run all the test cases on Crosswalk
platforms and browsers. So it should not include any test suites
specific for Tizen platform.


zqzhang:
Wait a minute for merging, let me check other suites.

PR #2497, opened 2015-07-15
[tools] New a directory to store the suite list

qiuzhong:
To maintain the test suite list on github, we need to add a new
release_suite_lists directory in crosswalk-test-suites/tools and put
all the test suite list files in this directory.

https://crosswalk-project.org/jira/browse/XWALK-4617


zqzhang:
Which script(s) will use these suite lists?

If I understand correctly, they are ONLY used by `tools/build/pack_wts.sh`. Why not put them into `tools/build/`? Or make a subdirectory there?

tools/release_suite_lists/beta13_list
```diff
@@ -0,0 +1,64 @@
+aAPKLIST="sampleapp-android-tests
```
zqzhang: Error here?

Please make each list in alphabetical order.

zqzhang: Are these lists targeting for branches of the repository? For example, this list `beta13_list` is for https://github.com/crosswalk-project/crosswalk-test-suite/tree/Crosswalk-13; `beta14_list` for Crosswalk-14; and `release_list` for the master branch?

zqzhang:
If these lists are targeting for branches, I would suggest we maintain at most 2 lists in each branch (and update them in need): one for weekly test suite release (`tools/build/suites/WEEKLY`), the other for web testing release (`tools/build/suites/WTS`, currently it is `tools/build/test_suites.list`).

qiuzhong:
OK. We'll split the list files to commit on different branches.

qiuzhong:
Will recommit these test suite list files.

PR #2498, opened 2015-07-15
[ATIP] Initial uiautomator as backend in ATIP

Honry:


Honry:
@yugang, could you please help to review it? And the verification is in progress.

Honry:
Since it contains too many updates and is too difficult to review, I'd like to separate it.

PR #2509, opened 2015-07-17
[ATIP] Rename debug files and add uiautomator env

Honry:
Rename webdriver_config to bdd_config to compatible with uiautomator
  Add uiautomator env in set_env.sh
  Rename json files


zqzhang:
> Rename webdriver_config to bdd_config to compatible with uiautomator

Does that mean `webdriver` is something a kind of `bdd`?

Honry:
```webdriver``` is just like a backend of ```bdd```, a way to implement ```bdd```.

zqzhang:
Merge as-is.

PR #2516, opened 2015-07-17
[embeddingapi] Add the pyunit structure for auto build

zhuyongyong:
- Add common methods that auto build will call
- Add demo code for auto build
- Add arch file to configure android device type
- Add crosswalk-type to configure the crosswalk is beta or stable or canary
- Add crosswalk-version to configure the checked crosswalk version
- Add libraries.gradle build.gradle for gradle build
- Add pom.xml for maven build

https://crosswalk-project.org/jira/browse/XWALK-4698


embeddingapi/embedding-build-android-tests/crosswalk-version
```diff
@@ -0,0 +1 @@
+14.43.343.20
```
zhuyongyong: I suddenly find it will be better to use https://github.com/crosswalk-project/crosswalk-test-suite/blob/master/VERSION instead of this crosswalk-version file.

jennycao:
LGTM

PR #2519, opened 2015-07-17
[usecase-ios]Update tests description for all ios tests

wanghongjuan:
- No check point that test project XWalkView build, so delete it

Impacted TCs num(approved): New 0, Update 4, Delete 1
Unit test Platform: Android
Android test result summary: Pass 4, Fail 0, Block 0


ibelem:
LGTM

PR #2521, opened 2015-07-20
[webapi] Remove bdd.config.xml for bdd tests

JianfengXu:
Due to Tizen platform has been removed, so remove the related config file for bdd tests.


zqzhang:
I am OK to live with this. How about you, @Honry ?

Honry:
But most of these tests can not run on Tizen since some steps in *.feature files only support on Android. And we have to create config.xml for new bdd-app tests to keep consistent, I suggest remove it.

PR #2524, opened 2015-07-20
Remove unnecessary configure files

haoxli:
- For android only, remove "config.xml,inst.wgt.py,inst.xpk.py" and update suite.json to keep android configure only.
- For tizen only, remove "inst.apk.py" and update suite.json to keep tizen configure only, manifest.json is used for xpk package.

BUG=https://crosswalk-project.org/jira/browse/XWALK-3375


haoxli:
Verified the package building.

zqzhang:
Good catch, thanks.

PR #2526, opened 2015-07-20
[ATIP] Move webdriver specfic steps from common to web

Honry:
And rename context dict from context.webs to context.apps to
 compatible with uiautomator
 BUG=https://crosswalk-project.org/jira/browse/XWALK-4650


zqzhang:
LGTM

PR #2531, opened 2015-07-21
[ATIP] Add new step in ATIP

JianfengXu:
BUG=https://crosswalk-project.org/jira/browse/XWALK-4640


tools/atip/atip/web/steps.py
```diff
@@ -337,3 +337,7 @@ def compare_two_value(context, first, second):
 @step(u'"{first}" should be less than "{second}" beyond "{value}"')
 def compare_two_value_with_range(context, first, second, value):
     assert context.web.compare_two_values_range(first, second, value)
+
+@step(u'"{value}" should be between "{first}" to "{second}"')
```
Honry: Where you get the "value"?  Shouldn't it be from some key?

Honry:
LGTM

PR #2555, opened 2015-07-23
Support json minimize in formater tool.

yugang:
@zqzhang , could you help to check? thanks

### DIFF
--- a/tools/format/json_formatter
+++ b/tools/format/json_formatter
@@ -154,6 +154,12 @@ def main():
             dest="symlinks",
             action="store_true",
             help="follow symlinks")
+        opts_parser.add_option(
+            "-m",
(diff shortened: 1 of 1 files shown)


tools/format/json_formatter
```diff
@@ -154,6 +154,12 @@ def main():
             dest="symlinks",
             action="store_true",
             help="follow symlinks")
+        opts_parser.add_option(
+            "-m",
+            "--minimize",
+            dest="minimize",
+            action="store_true",
+            help="minimize the json file")
```
zqzhang: `minimize JSON file to one line`?

yugang: not only one line, the new json will not include any unuseful char.

PR #2557, opened 2015-07-23
[tools] Put test author into metadata

zqzhang:
Historically test developer's name was put into file comments
immediately after license info. At the very beginning, this template is
OK for internal using.

As projects moving forward, these tests are expected to contribute to
upstream, e.g. W3C. It is important to put the author info into metadata
rather than into comment, because this comment including license and
author info must be removed before submission.

Sometimes, the authorship informations in metadata are used to
respect the authors, for example CSS WG nightly unstable lists all
authors in its acknowledgements section at
http://test.csswg.org/suites/css-transforms-1_dev/nightly-unstable/
Therefore both Intel and co-author Jieqiong Cui are acknowledged.

Sometimes, people can use this author metadata to talk test issues,
which is easier than finding author(s) via git log.

These templates are used for new test cases development; not for
existing tests.


zqzhang:
@jennycao @Honry @ibelem could you please review this? Any comment is welcome.

zqzhang:
CC @haoxli @chunywang @kangxu @JianfengXu

zqzhang:
All, I will also try to leverage some good parts of

* http://testthewebforward.org/docs/test-templates.html
* https://www.chromium.org/blink/serviceworker/testing

Especially the `Promises` and `Test Harness specific issues` in the 2nd one.

Stay tune for the templates update and thank you for all feedback in advance.

ibelem:
Up!

PR #2563, opened 2015-07-24
[wts] Fix wts gap for Canvas

kangxu:
There're still 4 tests gaps between wts and normal tests, it's
due to the following reasons:

- In wts, 2d.scaled-manual.png is treated as test

- In tests.xml, canvasgradient_addColorStop_INDEX_SIZE_ERR.html is
treated as one test, but there're 4 sub tests inside in fact.

Impacted tests(approved): new 0, update 110, delete 1
Unit test platform: WTS
Unit test result summary: pass 850, fail 20, block 0


zqzhang:
> In wts, 2d.scaled-manual.png is treated as test

This shall be a bug of WTS. @yugang @BruceDai

zqzhang:
Merge as-is.

PR #2566, opened 2015-07-24
[embeddingapi]Auto embeddingapi build cases

zhuyongyong:
Impacted tests(approved): new 1, update 6, delete 6
Unit test platform: Crosswalk Project for Android 14.43.343.20
Unit test result summary: pass 7, fail 0, block 0

BUG=https://crosswalk-project.org/jira/browse/XWALK-4698


zhuyongyong:
@jennycao I thought thus can directly use the built apk to install, launch, stop, uninstall. These apks are built in different ways, so need to install, launch, stop, uninstall to check if the apk is good. so install and so on is a check way.

PR #2567, opened 2015-07-24
[sample app] Update server IP and port for webrtc sample app

qiuzhong:
Currently, the server IP and port in the code of sample app webrtc
is invalid for the test team. So update them to a valid value for the
test team before packing the test suite.

https://crosswalk-project.org/jira/browse/XWALK-4663


ibelem:
LGTM

PR #2234, opened 2015-05-19
[webapi] Stop testing against W3C SysApps API specs

zqzhang:
Because the W3C SysApps WG and APIs are dead, I has proposed to stop
testing and to remove spec links from Crosswalk Project. The discussion
is still on going internally. Anyway, these test suites shall be in
sync with the final decision.

https://crosswalk-project.org/jira/browse/XWALK-48
https://crosswalk-project.org/jira/browse/XWALK-49
https://crosswalk-project.org/jira/browse/XWALK-53
https://crosswalk-project.org/jira/browse/XWALK-117
https://crosswalk-project.org/jira/browse/XWALK-568

PR #2576, opened 2015-07-27
[webapi] Stop testing against Messaging API

haoxli:
This has been discussed internally and gotten agreement.
See more at #2234.

https://crosswalk-project.org/jira/browse/XWALK-53


haoxli:
There is no usecase for SysApps Messaging API.

PR #2590, opened 2015-07-30
[webapi] Fix duplicated test purposes issue

kangxu:


kangxu:
@zqzhang  I will fix same issue in other suites  in this PR with several commits

kangxu:
@zqzhang  I have finished fix duplicated test purpose for test suites in webapi

PR #2599, opened 2015-07-31
[webapi] Add baseline data for Nexus3

JianfengXu:
All the baseline data are provided by LiJun, there is no Nexus3 device around me, so I can not provide UT.

BUG=https://crosswalk-project.org/jira/browse/XWALK-4678


Honry:
LGTM

PR #2609, opened 2015-08-04
[ATIP] Modify atip entry env variables acquire

jacky-young:
-Modify atip entry env variables acquire
-increase program compatibility


jacky-young:
@Honry please review it.

tools/atip/tests/environment_android.py
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) 2015 Intel Corporation.
```
Honry: Please rename it as environment.android.py

Honry:
LGTM

PR #2615, opened 2015-08-05
[misc] Add manual tests for System test

JianfengXu:
Failure Analyse: LocalStorage is not persistent, related bug : https://crosswalk-project.org/jira/browse/XWALK-4396. Cookie can not be set successfully on Crosswalk, need HTTP server.

Impacted TCs num(approved): New 33, Update 0, Delete 0
Unit test Platform: Crosswalk Project for Android 15.44.374
Unit test result summary: Pass 23, Fail 10, Blocked 0


ibelem:
@JianfengXu 
Crosswalk_FileWriter_FileReader_upgrade.html 
The expected result for upgrading is not "Repeat step 2 and step 3." then "This new content comes from FileWriter" display in textarea after clicking "write file". It should be able to read the file directly after upgrading app.

ibelem:
For failed cases, do we have a valid bug in JIRA?  Which cases are failed?

ibelem:
LGTM

PR #2617, opened 2015-08-05
[wrt] Fix duplicated test purpose issue

kangxu:
- Fix duplicated test purpose issue for internetstdmanu, rt
, rtcoremanu and sharedmodemanu


Honry:
LGTM

PR #2634, opened 2015-08-07
[atip-android] add uiautomator finding relative object steps

jacky-young:
- add uiautomator finding relative object steps


jacky-young:
@Honry please review it.

PR #2641, opened 2015-08-07
[webapi] Update xml to support subcase mode for Selectors Level2

kangxu:
Impacted tests(approved): new 2526, update 0, delete 0
Unit test platform: Crosswalk for Android 15.44.384
Unit result summary: pass 1737, fail 874, block 0

https://crosswalk-project.org/jira/browse/XWALK-4770

### DIFF
--- a/webapi/tct-selectorslevel2-w3c-tests/tests.full.xml
+++ b/webapi/tct-selectorslevel2-w3c-tests/tests.full.xml
@@ -9,9 +9,9 @@
         </description>
         <specs>
           <spec>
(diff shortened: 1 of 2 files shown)


zqzhang:
Though such kind of update makes the commit history ugly, we have to accept it firstly because separating them will be a waster of time.

@kangxu and @haoxli will make 2 new pull requests for the rest test suites to be updated: one is for the XML files format and the other is for real change. Therefore, merge this PR as-is.

PR #2650, opened 2015-08-10
[wrt] Update the inst script for securitymanu

wanghongjuan:
- Update the inst.apk.py for that can install all apk in this folder


ibelem:
@wanghongjuan Please add some comments about the update reasons for such PRs in the future.

ibelem:
LGTM

PR #2673, opened 2015-08-12
[atip-android] Add atip android execute command step

jacky-young:
-Add atip android execute command step

-Verified locally


jacky-young:
@Honry please help to review it.

Honry:
LGTM

PR #2701, opened 2015-08-19
[apptools-iOS] Do not allow user to name the project as 'www' for apptools iOS

yunxliu:
Do not allow user to name the project as "www"

Unit test platform: [Android]

https://crosswalk-project.org/jira/browse/XWALK-4813

### DIFF
--- a/apptools/apptools-ios-tests/apptools/allpairs.py
+++ b/apptools/apptools-ios-tests/apptools/allpairs.py
@@ -54,7 +54,8 @@ def generate_cmd():
         'org.123example.xwalk',
         'org.example._xwalk',
         'org.xwalk.Tests',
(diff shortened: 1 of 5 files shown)


Honry:
Please improve the commit msg as "Do not allow user to name the project as "www""

Honry:
Beside, I think all the backends should have this checkpoint.

yunxliu:
The reason why you cannot use www as the suffix of the app id is, it use 'www' directory to place the html contents internally, and use the suffix of app id as the iOS project directory name. If the suffix is 'www', then the two directories will be mixed up. It allow user to name the project as "www" on Android and Deepin, so don't need have this checkpoint.

Honry:
Got, LGTM

PR #2704, opened 2015-08-19
[usecase-embedding] add a usecase for 'onDraw' in both sync and Async

Shao-Feng:
Add a usecase for onDraw(Canvas canvas)

Unit test platform: [Android]
Unit test result summary: pass 0, fail 2, block 0

A bug is reported https://crosswalk-project.org/jira/browse/XWALK-4842


jennycao:
LGTM

PR #2718, opened 2015-08-21
[webapi] Add new cases for presentation api

chunywang:
- Presentation API is been implemented to follow Spec
http://www.w3.org/TR/2015/WD-presentation-api-20150701/
- NavigatorPresentation interface:
 -  getSession(), getSessions(), onsessionavailable event doesn't exist,
remote page receives message by "messagetoremote" event handler on Crosswalk
 - joinSession doesn't implement completely, we can't get a seeion object from navigator.presentation.joinSession
 -  we can't get a Availability typeof object but function from navigator.presentation.getAvailability
 - we can't get a PresentationSession object from navigator.presentation.startSession
- PresentationSession interface(Session function on Crosswalk)
 -  close(), onStatechange, binaryType doesn't exist
 -  the "state" value got false not "connected" or "disconnected"
 -  send Blob, ArrayBuffer doesn't exist

Impacted TCs num(approved): New 13, Update 0, Delete 15
Unit test Platform: Crosswalk Project for Android 16.45.407.0
Unit test result summary: Pass 3, Fail 6, Blocked 4

BUG=https://crosswalk-project.org/jira/browse/XWALK-5083


zqzhang:
Please refer to https://crosswalk-project.org/jira/browse/XWALK-4646, so that stakeholders can get testing update info.

zqzhang:
Merge as-is for further test development.

PR #2722, opened 2015-08-21
[webapi] Remove tests on HTMLMediaElement.controller for Video

kangxu:
As per https://w3c.github.io/test-results/html/complete-fails.html, no browser supports the HTMLMediaElement.controller attribute and MediaController interface well. So remove corresponding test cases, and add them back if upstream Chromium and Crosswalk support them.

Bakup in https://github.com/crosswalk-project/crosswalk-test-suite/tree/designed/webapi/tct-video-html5-tests

BUG=https://crosswalk-project.org/jira/browse/XWALK-1832

PR #2723, opened 2015-08-21
[webapi] Remove tests on HTMLMediaElement.controller for Audio

kangxu:
As per https://w3c.github.io/test-results/html/complete-fails.html, no browser supports the HTMLMediaElement.controller attribute and MediaController interface well. So remove corresponding test cases, and add them back if upstream Chromium and Crosswalk support them.

Bakup in https://github.com/crosswalk-project/crosswalk-test-suite/tree/designed/webapi/tct-audio-html5-tests

BUG=https://crosswalk-project.org/jira/browse/XWALK-1831


zqzhang:
See my comments at #2722

zqzhang:
Thanks, please also apply to crosswalk 15 branch.

PR #2739, opened 2015-08-25
[usecase-linux] Remove usecase for Media Codec

kangxu:
From now on Crosswalk Project only cares about building ffmpeg with
ffmpeg_component=shared_library, and then it is up to Deepin to replace
the shared library with whatever they want. So remove tests on proprietary
codecs required by feature XWALK-3705.

https://crosswalk-project.org/jira/browse/XWALK-4796


zqzhang:
Good catch. Please also apply to Crosswalk-15 branch, thanks.

PR #2742, opened 2015-08-25
[webapi] Update test type to auto for Gumallow manual tests

kangxu:
Dialog on video/audio stream didn't prompt in zte or memo, no need user interact. So change tests type from manual to auto. Automation these tests if dialog prompt again in later crosswalk release

Impacted tests(approved): new 0, update 14, delete 0
Unit test platform: Crosswalk for Android 16.44.395.0
Unit result summary: pass 17, fail 7, block 2

### DIFF
--- a/webapi/tct-gumallow-w3c-tests/tests.full.xml
+++ b/webapi/tct-gumallow-w3c-tests/tests.full.xml
@@ -9,7 +9,7 @@
         </description>
         <specs>
           <spec>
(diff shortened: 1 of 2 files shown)


zqzhang:
I don't think this update is acceptable.

zqzhang:
Close this as-is. If you think this shall be accepted, please reopen it with comment saying that this change has no impact on testing on other platforms/devices/OSs, thanks.

CC @Honry since this is something against automation.

Honry:
Let's keep eyes on https://crosswalk-project.org/jira/browse/XWALK-4907, we think this is a Crosswalk issue since a confirm prompt is expected action and Chrome 44 on Android does so.

PR #2743, opened 2015-08-25
[cordova] Update component name and category

zhuyongyong:
BUG=https://crosswalk-project.org/jira/browse/XWALK-4887


Honry:
LGTM, and please also back-port to Crosswalk-15 branch.

PR #2749, opened 2015-08-25
[webapi] Remove WebGL tests that fail to be submitted to upstream

chunywang:
- Most of the tests are checking existence, type, etc.
which shall be covered by existing WebGL conformance tests
- Removed 1275 tests each other for webgl and webgl-py

https://crosswalk-project.org/jira/browse/XWALK-4888


zqzhang:
Thank you for the PR, @chunywang 

However, we were asked to keep the WebGL test suite as previous release. So I would like to close this PR, and to accept #2750.

PR #2751, opened 2015-08-25
[usecase-embedding] Add a usecase for checking overScrollBy method

Shao-Feng:
Unit test platform: [Android]
Unit test result summary: pass 0, fail 2, block 0

Bug is reported
https://crosswalk-project.org/jira/browse/XWALK-4894


jennycao:
LGTM

PR #2760, opened 2015-08-26
[usecase] Add test for initialize

wanghongjuan:
Unit test Platform: Android
Android test result summary: Pass 1, Fail 0, Block 0

BUG=https://crosswalk-project.org/jira/browse/XWALK-4794


usecase/usecase-wrt-android-tests/samples/Initialize/res/videos.html
```diff
@@ -0,0 +1,14 @@
+﻿<!DOCTYPE html> 
+<!-- Copyright © 2015 Michele Placentino,Adriano Perla. All Rights Reserved-->
```
ibelem: +
@zqzhang Can we leverage this kind of sample as test cases directly in Crosswalk Test Suite?

zqzhang: No, if just this line. Could you please point me the origin of these files?

ibelem:
@zqzhang The original test case comes from http://www.filedropper.com/adriandemocanary004arm in https://crosswalk-project.org/jira/browse/XWALK-4778 . Any suggestions? I think we can cut down to simple test case rather than demo by ourselves.

@wanghongjuan

zqzhang:
> I think we can cut down to simple test case rather than demo by ourselves.

Yes, this is better.

By the way, I cannot access to http://www.filedropper.com/adriandemocanary004arm. It is blocked by Intel network.

wanghongjuan:
Updated

ibelem:
LGTM

PR #2764, opened 2015-08-26
[usecase-embedding] Add a usecase to check if onVisibilityChanged works

Shao-Feng:
Unit test platform: [Android]


jennycao:
LGTM

PR #2765, opened 2015-08-27
[usecase-embedding] remove the useless 'import' which refer a out-of-…

Shao-Feng:
[usecase-embedding] remove the useless 'import' which refer a out-of-scope library

build successfully with pack.sh


jennycao:
Add UT info in comments

PR #2775, opened 2015-08-31
[webapi] Sync notification tag same test to latest W3C

chunywang:
As per http://www.w3.org/TR/notifications/#showing-a-notification,
a notification in the list of pending notifications or the list of active notifications
whose tag equals notification's tag and whose origin equals notification's origin,
in the same position, close old and show new.

A example introduced http://www.w3.org/TR/notifications/#tags-example

Impacted TCs num(approved): New 0, Update 1, Delete 0
Unit test Platform: Crosswalk Project for Android 16.44.396.0
Unit test result summary: Pass 1, Fail 0, Blocked 0

BUG=https://crosswalk-project.org/jira/browse/XWALK-4841


zqzhang:
Thanks, please also apply to Crosswalk 15 branch.

PR #2795, opened 2015-09-06
[misc] Deprecate test cases for SysApps demos

zqzhang:
This depends on
https://github.com/crosswalk-project/crosswalk-demos/pull/50

Moreover QA has communicated with stakeholders that crosswalk-demos will
be out of testing scope because the crosswalk-samples has been
officially published in crosswalk-project.org


zqzhang:
@haoxli @jennycao PTAL

This PR may be conflicted by potential bug-fixing against https://crosswalk-project.org/jira/browse/XWALK-4975. So I think it is OK to accept it even the dependency PR isn't merged.

haoxli:
LGTM

PR #2405, opened 2015-06-26
[webapi] Integrate csswg tests for 2dtransforms

chunywang:
There are still 97 tests in designed, they need to be improved.

Impacted tests(approved): new 101, update 0, delete 30
Unit test platform: Crosswalk Project for Android 15.44.361.0
Unit test result summary: pass 95, fail 2, block 4

BUG=https://crosswalk-project.org/jira/browse/XWALK-4443

PR #2367, opened 2015-06-18
[webapi] Integrate csswg tests for 3dtransforms

chunywang:
There are still 58 tests in designed, they're lack of assertion description.

Impacted tests(approved): new 60, update 6, delete 7
Unit test platform: Crosswalk Project for Android 15.44.361.0
Unit test result summary: pass 66, fail 1, block 0

https://crosswalk-project.org/jira/browse/XWALK-4443

PR #2799, opened 2015-09-07
[webapi] Make README GitHub flavored markdown

zqzhang:
So that reading from browsing code gets better experience, e.g.
https://github.com/crosswalk-project/crosswalk-test-suite/tree/master/webapi/tct-sessionhistory-html5-tests


zqzhang:
@chunywang PTAL

zqzhang:
I have 2 findings when made these changes:

1. The CSS 2D/3D Transforms test suites can be combined. Historically these test suites were separated because there were huge number of test cases integrated from a test file submitted to W3C. This test file hasn't been accepted/approved, therefore we removed them from this test suite at #2367 and #2405.
2. The Fullscreen API is implemented based on https://fullscreen.spec.whatwg.org/, not the https://wiki.mozilla.org/index.php?title=Gecko:FullScreenAPI which once was supported by Tizen platforms.

CC @chunywang

PR #2810, opened 2015-09-09
[usecase-embedding] Add 4 use case for validating performLongClick an…

Shao-Feng:
…d requestFocus

Impacted TCs num(approved): New 4, Update 0, Delete 0
Unit test Platform: Android
Android test result summary: Pass 2, Fail 2, Block 0

Bug for LongClick listener is reported
https://crosswalk-project.org/jira/browse/XWALK-5018


jennycao:
LGTM

PR #2816, opened 2015-09-10
[Tools][Crosswalk-15] Update shared mode package method in webview plugin

zhuyongyong:
- Follow ﻿https://github.com/crosswalk-project/cordova-plugin-crosswalk-webview/blob/master/README.md
- Add one way to install webview plugin from npm
- It's successfull to pack cordova pkgs and all in one script can run successfully


Honry:
Merge as-is. For new way to install webview plugin from npm, since new npm plugin has not released yet, let's do verification by then.

PR #2850, opened 2015-09-16
[wrt] Apply JSON formatter to .json files for wrt

wanghongjuan:
- Make JSON default indent value as 2 spaces following Google JSON style guide.
- Apply JSON formatter to .json files for wrt
- python tools/format/json_formatter.py --recursive wrt/
- There are three .json files failed to format because they are manifest negative tests


Honry:
LGTM, @zqzhang, please help to review it.

PR #2857, opened 2015-09-17
[apptools-android] Update manual cases to test x86 and arm apk for apptools android

yunxliu:
- Update manual cases to test x86 and arm apk
- Add run the apk on android device step after building the apk

Impacted tests(approved): new 0, update 29, delete 3
Unit test platform: [Android]
Unit test result summary: pass 29, fail 0, block 0


Honry:
What about those five auto tests?

Honry:
LGTM

PR #2862, opened 2015-09-18
[doc] Fix more typos

zqzhang:
These typos shall be found when edit the text; but my Notepad++
doesn't work well on spell check.


zqzhang:
No need additional check, merging.

PR #2864, opened 2015-09-18
[wrt] Clean up tail space in README, feature, html and scripts

kangxu:
Clean up tail space in README, feature, html and scripts


Honry:
LGTM

PR #2865, opened 2015-09-18
Maintaince all-in-one packages in whitelist

qiuzhong:
For all-in-one packages, the packing will use a blacklist to ignore the
ones don't need to pack this kind of package.
Now we move the package list in PLATFORMS file and use a whitelist mechanism.

Related to: https://crosswalk-project.org/jira/browse/XWALK-4898

Signed-off-by: Zhong Qiu <zhongx.qiu@intel.com>


jennycao:
LGTM

PR #2871, opened 2015-09-21
[webapi] Remove Float32x4Array tests for SIMD

chunywang:
- SIMD spec: http://tc39.github.io/ecmascript_simd/#intro
- Crosswalk has removed SIMD Typed Array, SIMD.Float32x4Array, SIMD.Float64x2Array
and SIMD.Int32x4Array.

BUG=https://crosswalk-project.org/jira/browse/XWALK-5087


zqzhang:
CC @jennycao @yugang @ibelem 

This change will result in test case number difference between weekly testing and web testing service.

PR #2896, opened 2015-09-25
[atip] Update startswith and endswith to use innerHTML

haoxli:
For textarea element, innerText will be empty string when set its content,
it's better to use innerHTML for most elements.

Unit test platform: Crosswalk Project for Android 16.45.410.0

https://crosswalk-project.org/jira/browse/XWALK-4877


Honry:
LGTM

PR #2914, opened 2015-10-09
[apptools-android] Add tests to test offline operation for apptools android

yunxliu:
Unit test platform: [Android]

https://crosswalk-project.org/jira/browse/XWALK-5180

### DIFF
--- a/apptools/apptools-android-tests/apptools/create_offline.py
+++ b/apptools/apptools-android-tests/apptools/create_offline.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2015 Intel Corporation.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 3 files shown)


apptools/apptools-android-tests/apptools/create_offline.py
```diff
@@ -0,0 +1,57 @@
...
+#         Yun, Liu<yunx.liu@intel.com>
+
+import unittest
+import os
+import comm
+
+
+class TestCrosswalkApptoolsFunctions(unittest.TestCase):
+
+    def test_create_offline(self):
```
Honry: I don't see any offline operation.

yunxliu: Offline operation means that it allow to pass the name of a crosswalk zip directly, and not need to download the index pages from download.01.org. self.assertNotIn("Looking for", create_output), it doesn't looks up the download page at download.01.org to find out which version do download.

Honry:
LGTM

PR #2916, opened 2015-10-10
[webapi] Add reference baseline data for new devices

kangxu:


zqzhang:
@Honry is this necessary?

Honry:
Yes, but we need more time to do verification.

Honry:
I would like to close this PR since it is a big effort to maintain so many baselines.

PR #2945, opened 2015-10-15
[tools] Add result2csv for report convert

haoxli:
Add result2csv.py for convert result xml to csv offline.


zqzhang:
> tizen-qa-reports.sh.intel.com

Don't mention internal things.

zqzhang:
merge as-is

PR #2948, opened 2015-10-16
[apptools-iOS] Remove tests for apptools iOS

yunxliu:
Due to iOS only support fullscreen and standalone mode for manifest display field, so remove these 4 tests


Honry:
LGTM

PR #2970, opened 2015-10-22
Remove behavior tool test suite and documents

zqzhang:


zqzhang:
@jennycao

PR #2976, opened 2015-10-22
[usecase] Unify component for usecase test suite

haoxli:
-Unify component to "Crosswalk Use Cases/*"
-Update suite name

### DIFF
--- a/usecase/usecase-webapi-xwalk-tests/tests.android.xml
+++ b/usecase/usecase-webapi-xwalk-tests/tests.android.xml
@@ -1,145 +1,145 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet type="text/xsl" href="./testcase.xsl"?>
 <test_definition>
-  <suite name="Usecase WebAPI">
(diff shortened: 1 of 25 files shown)


usecase/usecase-webapi-xwalk-tests/tests.android.xml
```diff
@@ -3,143 +3,143 @@
 <test_definition>
   <suite name="Usecase WebAPI">
```
Honry: What about suite name?

haoxli: Yes, we also need update the suite name, but this will affect display name of title in usecase, this is why we change it. @jennycao What's your opinion?

jennycao: Yes, Pls also update the suite name

haoxli:
@zhuyongyong @wanghongjuan @yunxliu @jacky-young

jennycao:
ＬＧＴＭ

PR #2984, opened 2015-10-23
[pack-script] Remove pack script specific for WGT and XPK support

haoyunfeix:
BUG=https://crosswalk-project.org/jira/browse/XWALK-5273


zqzhang:
> BUG=XWALK-5273

Please don't use this because there are some other tasks in that ticket.

zqzhang:
Please update the commit message

PR #2992, opened 2015-10-26
[wrt] Remove Tizen related suites

Honry:
Remove Tizen related test suites because
Crosswalk Project has removed Tizen support at
crosswalk-project/crosswalk#3299


Honry:
@zqzhang @jennycao, PTAL.

PR #2997, opened 2015-10-27
Remove wgt,xpk config from suite.json

haoxli:
- Remove wgt, xpk config from suite.json
- Remove inst.xpk.py used for installing .xpk


Honry:
LGTM

PR #3024, opened 2015-11-03
[usecase] Change CSP test to manual

haoxli:
Unit test platform: Crosswalk Project for Android 17.45.424.0

BUG=https://crosswalk-project.org/jira/browse/XWALK-4600


zqzhang:
Why change back to manual? Only for PASS?

CC @Honry

haoxli:
CSP need use local server, we need to install docroot.zip to tinyweb.

Honry:
Can we make this as precondition?

haoxli:
It's ok for precondition, i also add it to precondition, but still need to setup precondition by manual.

zqzhang:
Merge as-is.

PR #3037, opened 2015-11-06
[webapi] Update flexiblebox tests

chunywang:
- justify-content property got pass result
- align-self got fail  when then flex item has no parent

Impacted TCs num(approved): New 2, Update 1, Delete 0
Unit test Platform: Crosswalk Project for Android 17.45.429.0
Unit test result summary: Pass 1, Fail 2, Blocked 0

https://crosswalk-project.org/jira/browse/XWALK-3227

### DIFF
--- a/webapi/tct-flexiblebox-css3-tests/flexiblebox/CSS3FlexBox_align-self.html
+++ b/webapi/tct-flexiblebox-css3-tests/flexiblebox/CSS3FlexBox_align-self.html
@@ -35,20 +35,47 @@ Authors:
   <head>
     <title>CSS3 Text Effects Test: CSS3FlexBox_align-self</title>
     <link rel="author" title="Intel" href="http://www.intel.com" />
-    <link rel="help" href="http://www.w3.org/TR/2012/WD-css3-flexbox-20120612/#align-self" />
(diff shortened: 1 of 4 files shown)


zqzhang:
> justify-content property got pass result

Could you please point me where is the checkpoint? Seems all the tests updated/created are for `align-self`.

chunywang:
1. Verify 'justify-content' property initial value.
    The property is applied to flex containers. (A flex container is the box generated by an element with a computed display of flex or inline-flex.)
https://github.com/crosswalk-project/crosswalk-test-suite/blob/master/webapi/tct-flexiblebox-css3-tests/flexiblebox/CSS3FlexBox_initial_value.html lack of the premise.

2. New created tests verify 'align-self' property. (A value of auto for align-self computes to the value of align-items on the element’s parent, or stretch if the element has no parent.)
  Crosswalk got a fail value when the the element has no parent.

zqzhang:
Merge as-is.

zqzhang:
@chunywang shall we apply this change to Crosswalk-16 branch to fix https://crosswalk-project.org/jira/browse/XWALK-3227?

chunywang:
OK, i will update it on Crosswalk-16.

PR #3061, opened 2015-11-20
Remove the .git directory to reduce the size

qiuzhong:
When packing cordova test suites, the .git direcotry of a cordova plugin
is offen included to the zip package. This may take a lot of disk space.
Remove the .git directory to reduce the size of cordova packages.

Related to: XWALK-5767


Honry:
LGTM

PR #3062, opened 2015-11-20
Update the correct .deb file path for deb packages

qiuzhong:
Update the correct file path of .deb files when packing the test suites
of Crosswalk for Linux.

Related to: XWALK-5783


jennycao:
LGTM

PR #3074, opened 2015-11-25
[tool][apk analyser] Add arm64 and x86_64 apps detection support

ibelem:
[Unit Test] Pass for arm64 and x86_64 apps test


ibelem:
@jennycao @zqzhang Please help to review the PR, thanks!

ibelem:
The reason why update the following line: External users may don't use crosswalk-test-suite at all when they just need to run the tool.

- <script src='../../resources/webrunner/jquery-1.10.2.min.js'></script>
+ <script src='https://code.jquery.com/jquery-2.1.3.min.js'></script>

tools/apkanalyser/result/apk-analyser-result.xsl
```diff
@@ -52,7 +56,7 @@
       #note, #dt { font-style: italic; margin-right: 6px; }
       .hili { color: rgba(211, 84, 0,1.0); }
     </style>
-    <script src='../../resources/webrunner/jquery-1.10.2.min.js'></script>
+    <script src='https://code.jquery.com/jquery-2.1.3.min.js'></script>
```
zqzhang: Though have no reason in commit message why make this change, I accept it as-is. Thanks.

PR #3117, opened 2015-12-09
[misc] Adjust test priority for Sample App

haoxli:
- pack, install, uninstall -> P1
- icon -> P2
- uninstall when running -> P3


zqzhang:
LGTM, thanks.

PR #3131, opened 2015-12-10
[Windows][hotfix] Remove the msi file after copy it to opt directory to avoid to get a wrong msi

zhuyongyong:
- It's fine to pack usecase-wrt-windows-tests


jennycao:
LGTM

PR #3132, opened 2015-12-10
[Crosswalk-17][Windows][hotfix] Remove the msi file after copy it to …

zhuyongyong:
…opt directory

- Remove the msi file after copy it to opt directory to avoid to get a wrong msi
- It's fine to pack usecase-wrt-windows-tests


jennycao:
LGTM

PR #3134, opened 2015-12-11
[usecase] Add all-in-one config for usecase docroot

haoxli:
https://crosswalk-project.org/jira/browse/XWALK-5905


zqzhang:
I have no idea the relationship between docroot and changes in this PR.

@Honry do you know this?

Honry:
CSP test requires a http server, we use tinyweb server via deploying source code in docroot.

jennycao:
LGTM

PR #3142, opened 2015-12-15
[Embedding-Extension] Modify manifest for new crosswalk-pkg script

jacky-young:


jacky-young:
@Shao-Feng please help to review.

PR #3174, opened 2015-12-23
[tools] Upload script to separate files per license

zqzhang:
Many thanks to @pjckcs for creating this script.

### DIFF
--- a/tools/codescan/separate.sh
+++ b/tools/codescan/separate.sh
@@ -1,0 +1,1002 @@
+#!/bin/bash
+##
+# This tool is used to seperate and group test files according to their license
+# for Code (IP) Scan.
+#
+# Copyright (c) 2015 Intel Corporation.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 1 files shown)


zqzhang:
@jennycao I would like to bring in this script as-is and then update it accordingly. What do you think?

zqzhang:
There is a license identification tool for Source Code named Ninka, written in Perl. Perhaps we can try it to identify the source files in this repository.

Usage:
```sh
# How do I use it?
ninka.pl <filename>

# How do I scan files in an entire package? Use xargs. For example, to scan all the files in a directory (and its children) use: 
find * | xargs -n1 -I@ ninka.pl '@'
```

See also
- http://ninka.turingmachine.org/
- https://github.com/dmgerman/ninka
- http://ninka.turingmachine.org/download/ninka-latest-version.tar.bz2

zqzhang:
Found one issue against this tool that it cannot correctly handle files with space in file names. Filed a bug at https://crosswalk-project.org/jira/browse/XWALK-6104

jennycao:
merge this script first for further improvement

PR #3178, opened 2015-12-25
[usecase-wrt]Update suite.json for packing

qiuzhong:
Add app-name property for all the subapp to avoid packing errors.
Fix a JSON format error.

Related to: https://crosswalk-project.org/jira/browse/XWALK-6058


qiuzhong:
@jennycao @haoxli, the fix is from the results of discussing with @haoyunfeix

Honry:
@qiuzhong, many thanks for fixing this, LGTM

PR #3179, opened 2015-12-25
[pack-script] Add pack support for 64bits

haoyunfeix:
1. Add x86_64 and arm64 options
2. Change service all-in-one packing path


zqzhang:
LGTM

PR #3228, opened 2016-01-11
[webapi] Update indexedDB tests to replace DOMError with DOMException

chunywang:
BUG=https://crosswalk-project.org/jira/browse/XWALK-5933


zqzhang:
Merge as-is for further improvement.

BTW, there are several bug fixing in W3C for Indexed DB, please take a look at them.

PR #3241, opened 2016-01-13
[Cordova][hotfix] Update the tests.full.xml

zhuyongyong:
- the num of tests.full.xml is less than tests.xml, so need to update the  tests.full.xml  to make it be consistent with the tests.xml.


Honry:
Please add more info about this update in your commit msg.

Honry:
LGTM

PR #3245, opened 2016-01-13
[usecase-wrt] Remove ChromeProtocol due to chrome protocol deprecated

haoxli:
chrome:// is not a required feature for crosswalk runtime.

https://crosswalk-project.org/jira/browse/XWALK-3595


Honry:
LGTM

PR #3271, opened 2016-01-18
[Cordova] Add 64 bit support for all in one pack script

zhuyongyong:
- It's fine to pack all in one packages

https://crosswalk-project.org/jira/browse/XWALK-5430


Honry:
LGTM

PR #3347, opened 2016-02-20
Fix building issue on app-tools, do not set arch option when building

haoyunfeix:
share mode apks.


Honry:
@haoyunfeix, please provide the UT result.

haoyunfeix:
Tested on embedded and shared mode.
72 packages built successfully.

Honry:
Thanks, @haoyunfeix , LGTM

PR #3364, opened 2016-02-25
[tools] Add git statistics script

haoxli:
./git_stat.py -r /path/to/crosswalk-test-suite --since 2016-1-1

BUG=https://crosswalk-project.org/jira/browse/XWALK-6119


jennycao:
Thanks for make it happen:)

PR #3388, opened 2016-03-03
[webapi] Remove ttwf-css-3d-polygon-cycle

haoxli:
rotate 3d cannot render a polygon cycle using the way in ttwf-css-3d-polygon-cycle,
this require all divs in same layer, this will be more complex.

BUG=https://crosswalk-project.org/jira/browse/XWALK-4594


zqzhang:
> rotate 3d cannot render a polygon cycle using the way in ttwf-css-3d-polygon-cycle,
> this require all divs in same layer, this will be more complex.

Can we improve them?

zqzhang:
Or rewrite them?

haoxli:
@zqzhang, Now we have no good way to rewrite it. In my knowledge,  i think it is impossible to make that, except for the method in reference.

PR #3398, opened 2016-03-04
[webapi] Remove apt-cmd-opt command on Windows

chunywang:
WebGL test suite can't build if the command exists, when removed it doesn't effect pass rate.


zqzhang:
There are about 20 JIRA tickets mentioned `--ignore-gpu-blacklist` at

https://crosswalk-project.org/jira/issues/?jql=text%20~%20%22%5C%22--ignore-gpu-blacklist%5C%22%22

Have you checked the Android/Linux platforms?

chunywang:
It's OK for Android platform.

PR #3430, opened 2016-03-10
[Windows] Drop the media capture related cases from windows test

zhuyongyong:
- Since media capture is not supported, remove the media capture related cases on windows test

BUG=https://crosswalk-project.org/jira/browse/XWALK-6511


jennycao:
LGTM

PR #3433, opened 2016-03-10
[usecase-apptools] Add tests to test manifest orientation field for windows apptools usecase

yunxliu:
Impacted tests(approved): new 8, update 0, delete 0
Unit test platform: [Windows]
Unit test result summary: pass 8, fail 0, block 0

BUG=https://crosswalk-project.org/jira/browse/XWALK-6500

### DIFF
--- a/usecase/usecase-apptools-tests/samples-windows/ManifestOrientation/res/any/index.html
+++ b/usecase/usecase-apptools-tests/samples-windows/ManifestOrientation/res/any/index.html
@@ -1,0 +1,65 @@
+<!DOCTYPE html>
+<!--
+Copyright (c) 2015 Intel Corporation.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+* Redistributions of works must retain the original copyright notice, this list
+  of conditions and the following disclaimer.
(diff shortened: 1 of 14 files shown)


Honry:
You must add hints for tester that:
Orientations (other than landscape primary) from apps on windows works only in:
- tablets
- tablet mode for 2 in 1 devices

and only when not docked and when not having multiple monitors attached. Also the platforms should be Windows 8 +.

usecase/usecase-apptools-tests/samples-windows/ManifestOrientation/res/any/index.html
```diff
@@ -0,0 +1,65 @@
...
+<script src="../../../../js/common.js"></script>
+<script src="../../../../js/tests.js"></script>
+
+<div id="header">
+  <h3 id="main_page_title"></h3>
+</div>
+<div class="content">
+  <h4>Precondition</h4>
+  <ol>
+    <li>Please test it on tablets on Windows 8+ 2-in-1 in tablet mode, the device is not docked and not having multiple monitors attached</li>
```
Honry: or 2-in-1 in tablet mode

Honry:
LGTM

PR #3438, opened 2016-03-11
[Cordova][hotfix] Update the component for webapi-appsecurity-external-tests

zhuyongyong:
- Use "WebAPI/EXTERNAL/App Security API" instead of "Cordova Features/AppSecurityApi" for the component because this suite is not only for cordova, but also for windows.


Honry:
LGTM

PR #3473, opened 2016-03-24
[SlientDownload] Modify tests.xml slient download mode cases name

jacky-young:


jacky-young:
@Shao-Feng @ibelem please help to merge.

PR #3482, opened 2016-03-25
Update the test case list to eliminate the gap of test suites

qiuzhong:
Update the test suite list and we'll release test suites packages based on the
list files on github.


Honry:
Thanks, LGTM

PR #3488, opened 2016-03-26
Correct the Windows-Platform

qiuzhong:
Since webapi/tct-mediacapture-w3c-tests already remove support for
Windows before last update the gap test suites, so remove this test suite
from the list of Windows.


Honry:
LGTM

PR #3493, opened 2016-03-29
[usecase-wrt] Copy jquery.js to sub app for manual packing

haoxli:
Tester need copy jquery.js to ApplicationlocalStorage/res/res/js/
manually everytime, let's copy it to source code when packing.

Unit test platform: Crosswalk Project for Android 19.48.503.0


Honry:
LGTM

PR #3498, opened 2016-03-29
[Cordova][hotfix] Fix the count issue

zhuyongyong:
- Remove the spec.json as it won't be used
- Update the category to "WebAPI/EXTERNAL"
- Remove the duplicated 'status="approved" type="compliance"' in test.full.xml


Honry:
LGTM

PR #3508, opened 2016-04-01
[usecase-webapi]Update the usecase IAP samples to support app-tools

qiuzhong:
Update the usecase IAP Samples so that we can packing these apks
via crosswalk-app-tools.

Unit test platform: Crosswalk for Android 18.48.477.10
Unit test result summary: pass 2, block 0

BUG=https://crosswalk-project.org/jira/browse/XWALK-6689


qiuzhong:
@Honry , please take a review for the late update.

Honry:
LGTM

PR #3548, opened 2016-04-18
[webapi] Remove fileapi test

chunywang:
It's an old feature in spec
https://www.w3.org/TR/2011/WD-FileAPI-20111020/#valid
New spec https://www.w3.org/TR/FileAPI/#dfn-createObjectURL
has been removed this feature.

BUG=https://crosswalk-project.org/jira/browse/XWALK-6742


Honry:
LGTM

PR #3562, opened 2016-04-26
[apptools] Stop setting the Length variables in the global scope.

rakuco:
This partially reverts commit 501e71ca ("[apptools] Add extension file
for apptools").

Setting variables such as `x86Length` in the global scope is wrong, as
they will not be available in the test methods unless one uses a
`global` statement. If that is not done, we end up with failures like
this:

    ======================================================================
    ERROR: test_target_arch (__main__.TestCrosswalkApptoolsFunctions)
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "crosswalk_pkg_basic.py", line 312, in test_target_arch
        x86Length = x86Length + 1
    UnboundLocalError: local variable 'x86Length' referenced before assignment

### DIFF
--- a/apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
+++ b/apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
@@ -38,11 +38,6 @@ import sys
 sys.path.append("../")
 import comm
 
-x86Length = 0
-x86_64Length = 0
-armLength = 0
-arm_64Length = 0
-apkLength = 0
 
 class TestCrosswalkApptoolsFunctions(unittest.TestCase):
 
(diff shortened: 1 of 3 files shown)


baleboy:
LGTM, thanks for fixing this

r0b5t4:
LGTM, thanks

Honry:
@rakuco, many thanks for the fixing, LGTM. I will merge it once @yunxliu done the verification.

apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
```diff
@@ -38,11 +38,6 @@
 sys.path.append("../")
 import comm
 
-x86Length = 0
-x86_64Length = 0
-armLength = 0
-arm_64Length = 0
-apkLength = 0
```
Honry: You lost "apkLength" variable.

haoyunfeix:
Merge the PR firstly, then I will submit another PR to add "apkLength"

rakuco:
> Merge the PR firstly, then I will submit another PR to add "apkLength"

I don't think you need to send follow-up PR: `apkLength` is already set in each test that uses it (see `test_path_absolute()`, for example).

haoyunfeix:
Thank you for reminding~ :)
Yes, but for other tests e.g. 
`test_target_arch` in `CI/crosswalk_pkg_basic.py`
`test_create_package_target_bit_crosswalkzip` in `create_package_target.py`
 `test_create_arm32 `in `crosswalk_app_targets.py`
Still lack of this `apkLength` variable, I submitted #3572 to fix it.
Anyway, before merge that PR, let us QA do a UT and CI test confirm.... :)

rakuco:
ok, thanks for being thorough!

PR #3563, opened 2016-04-26
[apptools] Fix all failing crosswalk_pkg_basic tests.

rakuco:
Tests such as `test_keep_project` were needlessly relying on app-tools'
output format to determine the directory where the build tree was
created. app-tools has changed its output months ago and the test was
failing. On Windows, this also depends on app-tools pull request #129 to
work.

Other tests were failing in 64-bits mode, either because comm.py guessed
the wrong file name for the 64-bits zip files, and several tests failed
there because by default app-tools will only produce 32-bit APKs and the
test were not passing `--targets` to it.

`invokeCrosswalkPkg()` was also added to make invoking crosswalk-pkg and
checking its return code easier. `comm.getstatusoutput()` was left
intact because other tests depend on it, but it definitely should be
rewritten.

### DIFF
--- a/apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
+++ b/apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
@@ -28,6 +28,8 @@
 # Authors:
 #         Yun, Liu<yunx.liu@intel.com>
 
+import re
+import subprocess
 import unittest
 import os
 import zipfile
@@ -45,6 +47,18 @@ arm_64Length = 0
 apkLength = 0
 
 class TestCrosswalkApptoolsFunctions(unittest.TestCase):
+    def invokeCrosswalkPkg(self, params=None):
+        params = params or []
+        cmdline = []
+        if comm.HOST_PREFIX:
+            cmdline.append(comm.HOST_PREFIX)
+        cmdline.extend([os.path.join(comm.PackTools, 'crosswalk-pkg')])
+        cmdline.extend(params)
+        process = subprocess.Popen(cmdline, stdout=subprocess.PIPE,
+                                   stderr=subprocess.STDOUT)
+        output, _ = process.communicate()
+        self.assertEquals(process.returncode, 0)
+        return output
 
     def test_path_absolute(self):
         comm.setUp()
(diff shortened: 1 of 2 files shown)


rakuco:
+@baleboy, @r0b5t4

r0b5t4:
looking good, thanks!

yunxliu:
<code>def test_path_absolute(self)</code> checkpoint: default is stable without --crosswalk, and app path support absolute path, if package command with --crosswalk, need remove 
<code>version = comm.check_crosswalk_version(self, "stable")</code>
I will add testcase to test default is stable without --crosswalk

<code>def test_target_arch(self)</code> checkpoint: pack apk with --targets="arm64-v8a x86_64" or --targets="armeabi-v7a x86". ”--targets=64“ and ”--targets=32“ have been covered in other tests

apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
```diff
@@ -45,16 +47,32 @@
...
-        cmd = comm.HOST_PREFIX + comm.PackTools + \
-            "crosswalk-pkg --platforms=android --android=" + comm.ANDROID_MODE + ' --targets="' + comm.BIT + '" ' + comm.ConstPath + "/../testapp/create_package_basic/"
-        (return_code, output) = comm.getstatusoutput(cmd)
+        output = self.invokeCrosswalkPkg([
+            '--platforms=android',
+            '--android=' + comm.ANDROID_MODE,
+            '--crosswalk=' + comm.crosswalkzip,
+            '--targets=' + comm.BIT,
+            comm.ConstPath + '/../testapp/create_package_basic/',
+        ])
```
yunxliu: def test_path_absolute(self) checkpoint: default is stable without --crosswalk, and app path support absolute path, if package command with --crosswalk, need remove 
version = comm.check_crosswalk_version(self, "stable")
I will add testcase to test default is stable without --crosswalk

apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
```diff
@@ -282,26 +319,30 @@ def test_keep_project(self):
...
-            cmd = comm.HOST_PREFIX + comm.PackTools + \
-                "crosswalk-pkg --platforms=android --android=" + comm.ANDROID_MODE + " --crosswalk=" + comm.crosswalkzip + ' --targets="armeabi-v7a x86" ' + comm.ConstPath + "/../testapp/create_package_basic/"
-        return_code = os.system(cmd)
+        self.invokeCrosswalkPkg([
+            '--platforms=android',
+            '--android=' + comm.ANDROID_MODE,
+            '--crosswalk=' + comm.crosswalkzip,
+            '--targets=' + comm.BIT,
+            comm.ConstPath + '/../testapp/create_package_basic/',
+        ])
```
yunxliu: def test_target_arch(self) checkpoint: pack apk with --targets="arm64-v8a x86_64" or --targets="armeabi-v7a x86". ”--targets=64“ and ”--targets=32“ have been covered in other tests

haoyunfeix:
LGTM

haoyunfeix:
And liuyun will submit another PR to remove cases blocked by known issue.

apptools/apptools-android-tests/apptools/CI/crosswalk_pkg_basic.py
rakuco: My bad, I didn't notice the command line had changed here.

PR #3574, opened 2016-04-29
[webapi][realsense] Update realsense tests

chunywang:
There are some changes with realsense v19.6.1:
- FaceError, SPError, DEPError are replaced with DOMException
- enumeration values be all lowercase, and that multiple words be separated using dashes

Impacted TCs num(approved): New 0, Update 33, Delete 0
Unit test Platform: Crosswalk Project for Windows 20.49.516.0
Unit test result summary: Pass 19, Fail 14, Blocked 0

BUG=https://crosswalk-project.org/jira/browse/CTS-603
BUG=https://crosswalk-project.org/jira/browse/CTS-602


chunywang:
The failure cases don't reject with DOMException but Error.

PR #3584, opened 2016-05-05
[Crosswalk-19][usecase-wrt]Fix incorrect execution_type in tests.xml

haoxli:
execution_type should be auto if it's a bdd/ref test.


Honry:
LGTM

PR #3585, opened 2016-05-05
[WebAPI] Update js-test-pre.js to support webgl test cases for wts.

BruceDai:
BUG=https://crosswalk-project.org/jira/browse/CTS-788

### DIFF
--- a/webapi/webapi-webgl-khronos-tests/webgl/khronos/resources/js-test-pre.js
+++ b/webapi/webapi-webgl-khronos-tests/webgl/khronos/resources/js-test-pre.js
@@ -20,6 +20,25 @@
 ** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 ** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 */
+var caseName = document.title;
+var subcaseIndex = 1;
+
+function KhronosTest(name) {
+  this.name = name;
+  this.status = null;
+  this.message = null;
+}
+
+var khronosTests = [];
+var khronosTestMsg = null;
+
+function Status() {
+  this.status = null;
+  this.message = null;
+}
+
+var statusObj = new Status();
+statusObj.status = 0;
 
 (function() {
   var testHarnessInitialized = false;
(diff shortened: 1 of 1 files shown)


BruceDai:
@zqzhang @renchenglei please review this PR, thanks.

webapi/webapi-webgl-khronos-tests/webgl/khronos/resources/js-test-pre.js
```diff
@@ -20,6 +20,25 @@
 ** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 ** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 */
+var caseName = document.title;
```
zqzhang: This will change the test case name, right?

webapi/webapi-webgl-khronos-tests/webgl/khronos/resources/js-test-pre.js
```diff
@@ -140,6 +168,14 @@ function escapeHTML(text)
 
 function testPassed(msg)
 {
+    if (msg !== "successfullyParsed is true") {
```
zqzhang: We don't need to count the sub-case numbers; we only need to record a pass or fail for the whole test page. Does this change indicate this pass/fail criteria?

webapi/webapi-webgl-khronos-tests/webgl/khronos/resources/js-test-pre.js
BruceDai: It's only for WTS. To save subcases' results into result json file, WTS uses document titile suffixing with '/' and index as the subcase's name.

webapi/webapi-webgl-khronos-tests/webgl/khronos/resources/js-test-pre.js
BruceDai: It's only for WTS. To save subcases' results into result json file, WTS uses document titile suffixing with '/' and index as the subcase's name. If the judgment statement "if (msg !== "successfullyParsed is true") " satisfied, the result of the whole test page is BLOCK.

PR #3589, opened 2016-05-10
[usecase-webapi] Fix a FingerPrint extension directory typo

qiuzhong:
[usecase-webapi] Fix a FingerPrint extension directory typo.
    
FingerPrint example is copied from demo-express, so keep the
extension directory same as manifest.json in demo-express.


qiuzhong:
@Honry

Honry:
LGTM

PR #3590, opened 2016-05-10
[usecase][realsense] Add Photo set usecase

chunywang:
Failure reason: photo.queryDepth() throw AbortError exception.

Unit test Platform: Crosswalk Project for Windows 20.49.516.0
Unit test result summary: Pass 0, Fail 1, Blocked 0

BUG=https://crosswalk-project.org/jira/browse/CTS-348


zqzhang:
> Failure reason: photo.queryDepth() throw AbortError exception

What should be thrown?

chunywang:
queryDepth() method should return the promise with the depth map of the photo, but i don't know why photo.queryDepth() throws AbortError exception in the usecase.

usecase/usecase-realsense-xwalk-tests/samples/PhotoSet/js/main.js
```diff
@@ -0,0 +1,209 @@
...
+  if(currentPhoto == null) {
+    $("#log").text("please capture photo first!");
+    return;
+  }
+  photoCapture.takePhoto()
+    .then(function(photo) {
+      return photo.queryDepth();
+    })
+    .then(function(image) {
+      return currentPhoto.setDepthImage(image);
```
zqzhang: What is the value of `currentPhoto` here by the way? Shall it be assigned with `photo`?

usecase/usecase-realsense-xwalk-tests/samples/PhotoSet/js/main.js
```diff
@@ -0,0 +1,209 @@
...
+    })
+    .catch(errorCallback);	
+}
+
+function setDepthImage() {
+  if(currentPhoto == null) {
+    $("#log").text("please capture photo first!");
+    return;
+  }
+  photoCapture.takePhoto()
```
zqzhang: Seems you've to call `photoCapture.takePhoto` before doing any other operations like `queryDepth()`, right? If yes, the logic is NOT correct.

See https://github.com/crosswalk-project/realsense-extensions-crosswalk/blob/master/sample/src/photo.js#L99, it just uses the photo captured to do query operations.

chunywang: Yes, if i only do queryDepth() operation, it can return successfully,
 if i call photoCapture.takePhoto() before doing queryDepth operation, then it will get AbortError exception.

zqzhang: Good finding! This shall be a bug actually.

It shall support both from the camera [Example 1] or loaded from somewhere [Example 2 or 3], per https://crosswalk-project.github.io/realsense-extensions-crosswalk/spec/depth-enabled-photography.html#access-photo-information

zqzhang:
Let me merge this as-is. Please help file a bug accordingly, @chunywang Thanks.

PR #3601, opened 2016-05-14
[usecase-cordova] Remove shared mode in DefaultCrosswalkVersion

haoxli:
For default crosswalk version for webview plugin, no need shared
mode testing, it will use local runtime lib verison. default version
should be testing in embedded mode only.

Failure Analysis:
XWALK 6879 Webview plugin default crosswalk version is not 19
even if crosswalk is 19 in maven2/org/xwalk/xwalk_core_library_beta.

Unit test platform: WebView Plugin 1.7.0


haoxli:
@haoyunfeix,  PTAL

haoyunfeix:
lgtm

PR #3617, opened 2016-05-20
Update assertions after app-tools change

r0b5t4:
Host setup check is now disabled by default in app-tools when not running
in an interactive terminal (APPTOOLS-303). So the build output changed,
and the string search asserts looking for which platform is build also
have to be updated.

BUG=CTS-1732


haoyunfeix:
lgtm

PR #3621, opened 2016-05-23
[Embeddingapi] TC dev for Setting xwalk:background color

zhuyongyong:
BUG=https://crosswalk-project.org/jira/browse/CTS-674


haoxli:
This feature won't be implemented, close it.

PR #3633, opened 2016-05-27
Update the usage information for --pkg-version option

qiuzhong:
--pkg-version options can be used to both specify a crosswalk version
number or the absolute path of a crosswalk binary.


qiuzhong:
@Honry

Honry:
Thanks for the commit, LGTM.

PR #3654, opened 2016-06-14
[Crosswalk-20][Embeddingapi] Improve tests.full.xml to support launch auto cases

zhuyongyong:
BUG=https://crosswalk-project.org/jira/browse/CTS-1773


haoxli:
LGTM

PR #3662, opened 2016-06-17
[Embeddingapi] Use diff package name for embedding-api-android-tests …

zhuyongyong:
…and embedding-asyncapi-android-tests

BUG=https://crosswalk-project.org/jira/browse/CTS-1772


Shao-Feng:
Pls also update the 'inst.py' for install and uninstall this packages, alsowise this change will break inst.py

zhuyongyong:
@Shao-Feng  you can refer to the file "inst.apk.py" for the update change.

Shao-Feng:
do you mean the inst.apk.py will be rename as inst.py in the generated zip file?

zhuyongyong:
Yes

PR #3670, opened 2016-06-28
Fixed the bug CTS-1786 Remove webdriver-w3c-tests.

kaixinjxq:
This test suite was used for testing webdriver implementation, but it is never used for a long time, so we can remove it and archive it in before Crosswalk-20 branch.

BUG=https://crosswalk-project.org/jira/browse/CTS-1786


Honry:
LGTM

PR #3671, opened 2016-06-28
[Crosswalk-20][Embeddingapi] Use diff package name for embedding-asyn…

zhuyongyong:
…capi-android-tests

BUG=https://crosswalk-project.org/jira/browse/CTS-1772


haoxli:
lgtm

PR #3683, opened 2016-07-06
Resolve crash problem caused by WebCLKernel_setArg_basic

renchenglei:
Unit test platform: Crosswalk Project for Android 20.50.533.0
Unit fail reason: sync the test case to latest. The test case fail because Crosswalk doesn't support WebCL completely.

BUG=https://crosswalk-project.org/jira/browse/CTS-1767

### DIFF
--- a/webapi/webapi-webcl-nonw3c-tests/webcl/khronos/conformance/functionalityTesting/buildingAndRunning/setArg_basic.html
+++ b/webapi/webapi-webcl-nonw3c-tests/webcl/khronos/conformance/functionalityTesting/buildingAndRunning/setArg_basic.html
@@ -45,7 +45,7 @@ var kernel_arg_values = [
     [256, 4294967295, 4294967296, 0, -1], // UInt
     [[0, 5000], [50, 10], [2147483647, 2147483647], [2147483648, 2147483648], [-2147483648, -2147483648], [-2147483649, -2147483649]], // Long
     [[0, 1898989], [90, 10], [4294967295, 4294967295], [4294967296, 4294967296], [0, 0], [-1, -1]], // ULong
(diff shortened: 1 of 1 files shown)


renchenglei:
@Honry PTAL, Thanks a lot!

Honry:
Do you have the failing reason?

webapi/webapi-webcl-nonw3c-tests/webcl/khronos/conformance/functionalityTesting/buildingAndRunning/setArg_basic.html
```diff
@@ -77,7 +77,7 @@
         var output = wtu.createBuffer(webCLContext, webcl.MEM_WRITE_ONLY, inputArray.BYTES_PER_ELEMENT * resultArray.length);
         wtu.setArg(kernel, 0, input);
         wtu.setArg(kernel, 1, output);
-        wtu.setArg(kernel, 2, new Uint32Array([inputArray.length]));
```
renchenglei: @Honry We can can create a Uint32Array in 3 ways:
`new Uint32Array( length );`
`new Uint32Array( array );`
`new Uint32Array( buffer, byteOffset, length );`
So, it should be `new Unit32Array([length])` but not `new Uint32Array([inputArray.length])`. It's the reason that causes the platform crash.
Since the upstream has some updates, I sync the test case to the latest. The other changes are the updates.

Honry: I meant the failing reason in your UT result.

Honry:
LGTM

PR #3689, opened 2016-07-13
Fix the same case id in webapi-webgl-khronos-tests

renchenglei:
BUG=https://crosswalk-project.org/jira/browse/CTS-1823


Honry:
LGTM

PR #3690, opened 2016-07-13
Fixed bug CTS-1744 TC update for Deprecate and remove some nonstandard

kaixinjxq:
BUG=https://crosswalk-project.org/jira/browse/CTS-1744
Impacted tests(approved): new 0, update 20, delete 1
Unit test platform: Crosswalk for Android 21.50.540.0
Unit test result summary: pass 14, fail 5, block 1
Unit test fail reason: Test result is the same as Upstream result
Unit test delete reason: RTCSdpError is not used anywhere in the spec anyway.


Honry:
LGTM

PR #3691, opened 2016-07-13
[WebAPI] Test Integration for Upstream Batterystatus

kaixinjxq:
BUG=https://crosswalk-project.org/jira/browse/CTS-309
Impacted tests(approved): new 11, update 4, delete 2
Unit test platform: Crosswalk for Android 21.50.540.0
Unit test result summary: pass 12, fail 3, block 0
Delete Reason: Split the two files into eight new test files.
Fail Reason: Test result is the same as Upstream result and https://bugs.chromium.org/p/chromium/issues/detail?id=401553


Honry:
LGTM

PR #3700, opened 2016-07-15
Add new step to fix incompatibility issue in clicking app from task m…

Honry:
…anager between Xiaomi Pad2 and other devices

  Unit test platform: [Xiaomi Pad2][Memo8]
  BUG=https://crosswalk-project.org/jira/browse/CTS-150


haoxli:
LGTM

PR #3702, opened 2016-07-15
[Cordova] Upgrade the value of minSdkVersion parameter from 14(default) to 16 when building cordova tests.

BruceDai:
BUG=https://crosswalk-project.org/jira/browse/CTS-1831

PR #3704, opened 2016-07-18
[Cordova] Upgrade the value of minSdkVersion parameter from 14(defaul…

qiuzhong:
…t) to 16 when building cordova tests. (#3702)

* [Cordova] Upgrade the value of minSdkVersion parameter from 14(default) to 16 when building cordova tests.
BUG=https://crosswalk-project.org/jira/browse/CTS-1831

* [Cordova] Upgrade the value of minSdkVersion parameter from 14(default) to 16 when running cordova tests.


qiuzhong:
Cherry-pick from the master branch modification.

PR #3707, opened 2016-07-18
[release] Fix cca build issue with --minSdkVersion=16

qiuzhong:
Fix the cca build that can't recoginzed --minSdkVersion=16.
The solution for cca build is to use option --android-minSdkVersion=16

BUG=https://crosswalk-project.org/jira/browse/XWALK-7101


Honry:
LGTM

PR #3712, opened 2016-07-19
[release] Fix cca build issue with --minSdkVersion=16

qiuzhong:
Fix the cca build that can't recoginzed --minSdkVersion=16.
The solution for cca build is to use option --android-minSdkVersion=16

BUG=https://crosswalk-project.org/jira/browse/XWALK-7101


qiuzhong:
Cca applications (CIRC and Eh) can be built successfully after adding this option now.

Honry:
LGTM

PR #3729, opened 2016-07-22
Correct Meego medias origins and license

Honry:
- Update separate.sh with some improvement

 BUG=https://crosswalk-project.org/jira/browse/CTS-1804


Honry:
@haoxli PTAL. Thanks.

haoxli:
LGTM

PR #3679, opened 2016-07-05
Fixed bug CTS-1806 TC dev for RTCPeerConnection for Chromium 51

kaixinjxq:
BUG=https://crosswalk-project.org/jira/browse/CTS-1806
Impacted tests(approved): new 1, update 0, delete 5
Unit test platform: Crosswalk for Android 21.51.544.0
Unit test result summary: pass 1, fail 0, block 0

PR #3739, opened 2016-07-25
[WebAPI]Add promise-based versions of RTCPeerConnection methods and remove some nonstandard.

kaixinjxq:
BUG=https://crosswalk-project.org/jira/browse/CTS-1826
Impacted tests(approved): new 5, update 21, delete 1
Unit test platform: Crosswalk for Android 21.51.544.0
Unit test result summary: pass 22, fail 3, block 1
Unit test fail and block reason: Test result is the same as Upstream result.
Unit test delete reason: RTCSdpError is not used anywhere in the spec anyway.
New cases reason: Remove the five cases in https://github.com/crosswalk-project/crosswalk-test-suite/pull/3679, add again and update them in form of need.


Honry:
Intrusive commit, `	[Cordova] Upgrade the value of minSdkVersion parameter from 14...`

PR #3751, opened 2016-08-01
[usecase] Add the preference element with LoadUrlTimeoutValue attribute to config.xml

BruceDai:
set its' value be 600000 (10 minutes).
BUG=https://crosswalk-project.org/jira/browse/CTS-1855


BruceDai:
@haoxli PTAL

haoxli:
Please add the UT result in commit message.

PR #3758, opened 2016-08-03
[usecase] Add the preference element with LoadUrlTimeoutValue attribute to config.xml [Cordova] Remove minSdkVersion parameter.

BruceDai:
@haoxli PTAL


haoxli:
lgtm

PR #3767, opened 2016-08-10
[WebAPI] TC dev for AudioParams nominal range for Chromium 52

renchenglei:
Unit test platform: Crosswalk for Android 22.52.557.0

https://crosswalk-project.org/jira/browse/CTS-1862


renchenglei:
@Honry PTAL, Thanks a lot!

Honry:
Do you have the failure reason?

renchenglei:
@Honry The upstream also has the same error. :)

Honry:
Is there a upstream bug?

renchenglei:
@Honry Maybe it is influenced by the device. I will double check the fail.

PR #3769, opened 2016-08-17
[Embeddingapi] Correct subcase number in v7.XWalkSettingTest

haoxli:
There should be 11 subcases in v7.XWalkSettingTest(7 new tests
 added by commit fb02b606f52c242ea9a9c202205734935d5801b4)

https://crosswalk-project.org/jira/browse/CTS-1819


Honry:
LGTM

PR #3789, opened 2016-09-20
Improve inst.py for IoT platform

qiuzhong:
Update:
1. Use abspath when scp packages
2. Add a new inst.py for docroot on IoT


qiuzhong:
@Honry , requested by @BruceDai and had been tested.

Honry:
LGTM

PR #3790, opened 2016-09-20
Cherry-pick inst.py improvement from master to Crosswalk-20

qiuzhong:
Update:
1. Use abspath when scp packages
2. Add a new inst.py for docroot on IoT


Honry:
LGTM

PR #3812, opened 2016-10-10
Add the Crosswalk test suite list for IoT platform.

qiuzhong:


tools/build/released_suites/IoT-Platform
```diff
@@ -0,0 +1,77 @@
+IOTTESTSUITELIST="
```
Honry: What about the docroot test suite list?

Honry:
LGTM

PR #3819, opened 2016-10-12
Add the input color checkpoint.

kaixinjxq:
BUG=https://crosswalk-project.org/jira/browse/CTS-1905
Add the checkpoint about the input color.
It is fail in Crosswalk 21.51.546.7.
There is a feature bug XWALK-7367 to track it.


Honry:
LGTM

PR #3823, opened 2016-10-17
Change upload.txt to upload.doc in the checkpoint of input capture

kaixinjxq:
Update one case, the test result is pass on 22.52.561.4.


Honry:
LGTM

PR #3828, opened 2016-10-21
Add decode mp3 mp4 wav by decodeAudioData method

kaixinjxq:
Add three cases and update one, two cases can pass and two is fail, the fail cases have bug  to track.


Honry:
LGTM

PR #3832, opened 2016-10-26
Fixed the bug that Fail to get correct the 'form' property for a labe…

kaixinjxq:
…l in forms
BUG=https://crosswalk-project.org/jira/browse/CTS-1910
Impacted tests(approved): new 2, update 7, delete 0
Unit test platform: Crosswalk for Android 23.53.588.0
Unit test result summary: pass 8, fail 1, block 0
Test result is the same as Upstream result


Honry:
Please update the COPYING file as well.

Honry:
And also the tests.xml as you have two new tests added.

Honry:
LGTM

PR #3836, opened 2016-11-03
[cca]Add the cordova-platform@6.0.0 for cca

qiuzhong:
Add cordova-android 6.0.0 explicitly for cca build.

Unit test platform: Crosswalk Project for Android 23.53.589.0

BUG=https://crosswalk-project.org/jira/browse/CTS-1916


qiuzhong:
@haoxli , @Honry

Honry:
But what about testing the newer version in future?

qiuzhong:
Add a text to store the cordova-android version?

Honry:
Can we expose a param for specific version of cordova-android?

qiuzhong:
Yes. I think that's a good way.

qiuzhong:
Updated, PTAL

cordova/cordova-sampleapp-android-tests/sampleapp/comm.py
```diff
@@ -181,7 +181,7 @@ def create(appname, pkgname, mode, sourcecodepath, replace_index_list, self, ext
         do_remove(glob.glob(os.path.join(project_root, "www")))
         do_copy(sourcecodepath, os.path.join(tool_path, appname, "www"))
 
-def buildGoogleApp(appname, sourcecodepath, self):
+def buildGoogleApp(appname, sourcecodepath, self, cordova_android_ver = "6.0.0"):
```
Honry: I don't think make default value "6.0.0" is a good point, just ignore it. And pass specific version in the param.

qiuzhong: That will require the calling of this function more than 1 places.

qiuzhong:
@Honry , updated

Honry:
@haoxli, please also take a look.

haoxli:
Prefer to add cordova-android version following the version of cordova-plugin-crosswalk-webview, not fix it as 6.0.0.

